PR #9, opened 2022-12-09
Add pdm scripts for server dev

sleepycat:
This commit adds some helper scripts to make server developement nicer. Essentially `pdm db`, `pdm migrate` and `pdm start` all do roughly what you might expect, with the twist that the database is running postgres locally with docker. When you want to shut it down you can run `pdm stop`.

This uses 2 env files... `server.env` for the server and `postgres.env` for the database. Details are in the README.

### DIFF
--- a/server/README.md
+++ b/server/README.md
@@ -2,6 +2,52 @@
 
 This is a Django based API running GraphQL.
 
+## Running the server
+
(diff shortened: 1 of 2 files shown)


szeckirjr (approved):
LGTM!

PR #10, opened 2022-12-09
README Changes

szeckirjr:
Do these steps make sense?


sleepycat (approved):
Late to the party, but this looks good to me!

PR #11, opened 2022-12-14
Remix docker/docker-compose setup

sleepycat:
This commit is the first setup in readying cpho for Kubernetes. There was some good stuff happening before, but the idea here is to start to embrace the possibilities that pdm and kubernetes bring.
* Using bind mounts so that the frontend's hot module reloading works.
* Migrations are now a separate step, so that they can eventually be run as an initContainer.
* The entrypoint.sh file that ran migrations was no longer needed.
* Server related .env files now live in the server folder, more in keeping with
* the monorepo philosophy.
* The servers Dockerfile has been rewritten to work with lighter-weight images, PEP 582 style dependencies and running a single process with waitress. The idea being to let Kubernetes handle running multiple copies if needed.


tomcdona:
@sleepycat @szeckirjr @najsaqib have we figured out where this will be living in production? Are we going to be able to host this fully in GCP and have InfoBase read from it? Or are we on the hook to be in the HC/PHAC Tenant?

If we have to be in the tenant, we should be aligning with the ITAP app service deployments instead of kubernetes...
Let me know if we should chat about it :)

server/Dockerfile
```diff
@@ -1,52 +1,44 @@
...
-# Set the TMP variable to move pytest_cache outside of our working directory
-ENV TMP=/tmp
-
-# run entrypoint.sh
-ENTRYPOINT ["/server/entrypoint.sh"]
-
-# During debugging, this entry point will be overridden. For more information, please refer to https://aka.ms/vscode-docker-python-debug
-CMD ["python", "manage.py", "runserver", "0.0.0.0:8000"]
+# Build stage
+FROM python:3-alpine AS builder
```
tomcdona, line 2: 🙌 multi-stage builds FTW!

docker-compose.yml
```diff
@@ -1,81 +1,44 @@
...
+    working_dir: /server
+    env_file: ./server/server.env
+    # restart: on-failure
+    ports:
+      - 8000:8000
+    volumes:
+    - type: bind
+      source: ./server
+      target: /server
+    command: python manage.py runserver 0.0.0.0:8000
```
tomcdona, line 31: this is still the Dev edition of the workflow, there should be some sort of reverse proxy in front of this instead of the runserver... @sleepycat we're using an nginx container in front of the django container for ITAP, open to other ideas :)

tomcdona (changes requested):

najsaqib:
> @sleepycat @szeckirjr @najsaqib have we figured out where this will be living in production? Are we going to be able to host this fully in GCP and have InfoBase read from it? Or are we on the hook to be in the HC/PHAC Tenant?
> 
> If we have to be in the tenant, we should be aligning with the ITAP app service deployments instead of kubernetes... Let me know if we should chat about it :)

We are still keeping both options open. I think the intention is to go with whatever option allows us to go into production first.

And ITAP app service deployments isn't using Kubernetes eh? We can hit you up or maybe we can talk about what that will entail at the Action plan for Django meeting next week?

tomcdona:
[VPN-Needed](https://github.hc-sc.gc.ca/hs/cct-appserviceplan)

This is the terraform that put the wastewater functionapp into dev, all the django stacks will be going into an app service.
[ADO Work Item](https://dev.azure.com/HCPHAC/National%20Surveillance%20Project/_workitems/edit/3237) has more details.

Things are open for the deployment path for this, as well as ITAP/WW, worth more discussion. I'll try to add some tags and mentions in a few places over the coming days

docker-compose.yml
sleepycat: This is definitely a first pass aimed at getting a decent dev workflow going. More to follow!

szeckirjr (approved):
LGTM

tomcdona (approved):
cool-beans!
let's see what happens next :)

PR #14, opened 2023-01-31
Remove reference to frontend files from server

sleepycat:
This commit removes references to the frontends static files from the API. The two services are already separate, and this config is clearly left over from some experiment from days gone by. One of the sacred tenants of microservices is that they are independently deployable, and it's good to remove these to keep the system boundaries clear.


szeckirjr (approved):
LGTM :shipit: been meaning to do this for a while to remove the warnings

PR #15, opened 2023-02-01
Frontend Testing

szeckirjr:
# Frontend testing

Adds testing to React client using Jest, sets up the ts config and commands necessary, as well as give an initial idea of tests to be written with and without routing and GraphQL.
It also adds the run test steps to the docker file so it is included as part of the CI pipeline.


frontend/package.json
```diff
@@ -12,7 +12,6 @@
     "@testing-library/jest-dom": "^5.16.4",
     "@testing-library/react": "^13.3.0",
     "@testing-library/user-event": "^14.2.1",
```
sleepycat, line 14: These should be moved into devDependencies.

frontend/package.json
```diff
@@ -48,5 +46,11 @@
       "last 1 firefox version",
       "last 1 safari version"
     ]
+  },
+  "devDependencies": {
+    "@types/jest": "^29.4.0",
+    "jest": "^27.5.1",
```
sleepycat, line 52: ts-jest seems very upset by this version of jest. `"jest": "^29.4.2",` makes it happy.

sleepycat (changes requested):
npm install fails currently because `ts-jest` hates this version of jest. Bumping the version and moving those testing related libraries from dependencies to devDependencies would be perfect! This is an awesome first step on the testing journey!

sleepycat (approved):
:tada:

PR #16, opened 2023-02-07
Add basic hello world test for server

sleepycat:
This is a small step forward to get us started with testing the server.


szeckirjr:
sweet! should we add this to dockerfile to incorporate testing into CI?

sleepycat:
Exactly! I figured we'd get some of the basics in place and then tackle that stuff in follow up PRs... as well as steadily expanding the test coverage (which is clearly just "proof of life" level right now).

szeckirjr (approved):
lgtm 🚀 just setting up the testing script

PR #20, opened 2023-02-16
Add api route to envoy proxy config

sleepycat:
This commit ensures that paths prefixed with /api are forwarded to the server.


szeckirjr:
sweet! hopefully this is a temporary thing until file import/export gets migrated to GraphQL :) thanks for fixing it

szeckirjr (approved):

PR #27, opened 2023-03-16
Add wordmark component

LilaKelland:
Replaced static canada.svg with a [Wordmark component](https://github.com/PHACDataHub/cpho-phase2/blob/add-wordmark-component/frontend/src/components/atoms/Wordmark.tsx) containing inline SVG (outlined https://github.com/PHACDataHub/safe-inputs/issues/109). Worked with @daneroo who refactored the component & tests to limit colour options to only the 4 valid combinations listed [here](https://github.com/PHACDataHub/cpho-phase2/compare/canada.ca/en/treasury-board-secretariat/services/government-communications/design-standard/colour-design-standard-fip.html#toc1). Updated react-testing-library. Tests are [here](https://github.com/PHACDataHub/cpho-phase2/tree/add-wordmark-component/frontend/src/components/atoms/__tests__). Happy to move them if they should living in a different location.


szeckirjr (approved):
LGTM

PR #30, opened 2023-03-21
Fix local run

szeckirjr:
This PR fixes the ability to run both the frontend and backend locally (without Docker).
Since the Docker container assumes both services in the same port, the uri for the GraphQL client needs to be adjusted to account for the local run.


Stephen-ONeil:
Made irrelevant by new Django monolith structure.

PR #39, opened 2023-05-11
Add phac signature

samiatoui:
Created the PHAC Signature component. 

If no language is selected, the signature will appear in English and French, otherwise will appear based on the language prop selected.

Language selector does not exist yet so have to hard code the language.


Stephen-ONeil:
Closing as this PR has been invalidated by the move to a Django monolith approach, see #49 for reminder to revisit this with a Django-centric solution.

PR #42, opened 2023-06-12
Indicator CSV upload

msarar:
Indicator data files can now be uploaded.
there is a help modal that displays expected column names and column values.


server/cpho/jinja2/indicators/upload_indicator.jinja2
```diff
@@ -0,0 +1,24 @@
+{% extends 'base.jinja2' %}
+{% block content %}
+  <form action="{{ url('upload_indicator') }}"
```
AlexCLeduc, line 3: render form errors here, i think `{{ form.errors }}` will be good enough to develop, you can style it later by accessing it differently

server/cpho/views/indicators.py
```diff
@@ -48,6 +60,17 @@ class Meta:
...
+class UploadForm(forms.Form):
+    csv_file = forms.FileField(
+        required=True,
+        widget=forms.FileInput(
+            attrs={
+                "class": "form-control",
+            }
+        ),
+    )
+
```
AlexCLeduc, line 72: move the validation stuff from your form_valid into a `def clean_csv_file()` here. You can raise ValidationError rather than use messages

server/cpho/views/indicators.py
```diff
@@ -95,3 +118,259 @@ def get_context_data(self, **kwargs):
...
+
+class UploadIndicator(FormView):
+    template_name = "indicators/upload_indicator.jinja2"
+    form_class = UploadForm
+    # success_url = "/upload_indicators/"
+
+    def get_success_url(self):
+        return reverse("list_indicators")
+
```
AlexCLeduc, line 130: you probably won't need it at all, but you can override form_invalid if you want to use something like messages

server/cpho/views/indicators.py
```diff
@@ -95,3 +118,259 @@ def get_context_data(self, **kwargs):
...
+                data_dict.append(data_row)
+
+            messages.success(self.request, tdt("Data Uploaded Successfully"))
+
+        except Exception as err:
+            messages.error(
+                self.request, tdt("Unable to upload file: ") + repr(err)
+            )
+
+    def deduce_dimension_type(self, record, csv_line_number):
```
AlexCLeduc, line 199: even this can go in the form I guess? Ideally the form has all the logic, the view is just there to call form.save, message and redirect

server/cpho/views/indicators.py
```diff
@@ -48,6 +60,17 @@ class Meta:
     )
 
 
+class UploadForm(forms.Form):
```
AlexCLeduc, line 63: perhaps move this into an upload_indicators.py module so you can be looser with util functions like deduce dimension

server/README.md
```diff
@@ -95,20 +95,21 @@ note: recommended to use git bash inside vscode for all of this
 4. create a virtual environment in repo root (python -m venv venv)
 5. activate virtual environment (source venv/Scripts/activate or venv/bin/activate)
 6. install dependencies (`pip install -r requirements.txt -r requirements_dev.txt`)
-7. setting postgres:
+7. Start the db server: pg_ctl -D ~/pg/data/ -l logfile start
```
AlexCLeduc, line 98: Can we just add "(if applicable)" here? Only people who used the other guide for docker-less phac-windows will need to do this, but everyone will have to run the other commands

Thanks for updating the docs!

server/cpho/views/upload_indicators.py
```diff
@@ -0,0 +1,437 @@
...
+        "Province_AB": DimensionValue.objects.get(
+            dimension_type__code="province", value="ab"
+        ),
+        "Province_SK": DimensionValue.objects.get(
+            dimension_type__code="province", value="sk"
+        ),
+        "Province_MB": DimensionValue.objects.get(
+            dimension_type__code="province", value="mb"
+        ),
+        "Province_BC": DimensionValue.objects.get(
```
AlexCLeduc, line 206: Top level queries fail. For instance I can't even migrate or run any manage.py commands because these queries might fail. It's also an N+1 thing. Can we just refer to the codes, or use Q objects, e.g. `Q(dimension_type__code="province", value="sk")`

AlexCLeduc: Can these also be in a separate dict? They're kind of annoying to deal with inside the class, it's a lot of properties to collapse when trying to work with the class :)

server/cpho/views/upload_indicators.py
```diff
@@ -0,0 +1,437 @@
...
+            dim_val = None
+            lit_dim_val = None
+            if datum["Dimension_Type"] != "Age Group":
+                dim_val = non_literal_dimension_value_mapper[
+                    f'{datum["Dimension_Type"]}_{datum["Dimension_Value"]}'
+                ]
+            else:
+                lit_dim_val = datum["Dimension_Value"]
+
+            indData_obj, id_created = IndicatorDatum.objects.get_or_create(
```
AlexCLeduc, line 287: get_or_create will always create a new record whenever there's even a single new value. What if they want to update an existing value?

We can use the (indicator, dimension_type, dimension_value, literal_dimension_val) to identify an existing row. If it exists, write to it, otherwise create a new record.

AlexCLeduc (approved):

PR #43, opened 2023-06-16
WIP: dynamic age group

msarar:


server/cpho/views/indicator_data.py
```diff
@@ -133,21 +145,52 @@ def formset(self):
...
+                record = existing_data_by_literal_value[
+                    str(nlpv.literal_dimension_val)
+                ]
+            else:
+                print("not found")
+                record = IndicatorDatum(
+                    indicator=self.indicator,
+                    dimension_value=None,
+                    dimension_type=self.dimension_type,
+                    literal_dimension_val=nlpv.literal_dimension_val,
```
AlexCLeduc, line 182: I would think this block never gets called, right?

msarar: i think you're right!
ill comment out the else block and test it
:)

PR #46, opened 2023-06-29
Add .vscode/launch.json for VSCode debugging shortcuts

Stephen-ONeil:
Adds `runserver` and `test` items to the VSCode debugging menu. I always find this useful on django projects, maybe worth doing on other projects and in the template repo (unless y'all really prefer debugging outside of VSCode).

image

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,29 @@
+{
+"version": "0.2.0",
+  "configurations": [
+    {
+      "name": "runserver",
+      "type": "python",
+      "request": "launch",
+      "program": "${workspaceFolder}/server/manage.py",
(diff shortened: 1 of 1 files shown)


AlexCLeduc (approved):
Cool, I wonder if it works in windows and whether it'll do it in git-bash vs powershell

Stephen-ONeil:
> Cool, I wonder if it works in windows and whether it'll do it in git-bash vs powershell

Good question. Looks like VSCode on windows might default to powershell, but can be configured to use any other shell it can find on the system. Worth checking with one of the windows based devs, might be a good dev env set up tip (or maybe it can even go in the committed settings file).

PR #48, opened 2023-06-29
Fix test_create_indicator post data

Stephen-ONeil:
Trivial, although the `CreateIndicator` view/form seems to be WIP (no validation, etc) so I expect the shape of the data in this test will change again some time, right? Worth adding tests for the rest of the CRUD steps now?


AlexCLeduc (approved):
thanks!

AlexCLeduc:
Not sure if it's worth adding more tests just yet, I am adding periods and that is messing with all the parameters so we can hold off in the immediate term

PR #51, opened 2023-06-30
set up foundation for periods-based reporting

AlexCLeduc:
UI/UX flow is still a bit uncertain but we now have things we can tweak
commits 2,3 and 4 are unrelated

We now have calendar years, fiscal years and fiscal quarters. The input views are all tied to a single period.


server/cpho/models/indicators.py
```diff
@@ -65,6 +65,12 @@ def __str__(self):
 @add_to_admin
 @track_versions_with_editor_and_approval
 class IndicatorDatum(models.Model):
+    class Meta:
+        unique_together = [
+            ("indicator", "period", "dimension_type", "dimension_value"),
+            ("indicator", "period", "dimension_type", "literal_dimension_val"),
```
AlexCLeduc, line 71: Note these new constraints, see test_models.py for example invalid data

msarar (approved):
Looks good to me

PR #53, opened 2023-07-07
added Dependabot config

LilaKelland:
Added dependabot.yml. Have copied format and times from the [tracker](https://github.com/canada-ca/tracker/blob/master/.github/dependabot.yml#L1-L8) project as I imagine there was a reason for daily 11 am checks. 

Noticed there are a number of configurations we can set while reading through the docs. 
* Do we want to be assigning someone for these PRs?
* How do we want to handle version control between helper functions and this project while maintaining bug fixes for security purposes?

### DIFF
--- a/server/.github/dependabot.yml
+++ b/server/.github/dependabot.yml
@@ -1,0 +1,52 @@
+version: 2
+updates:
+# Dockerfile
+- package-ecosystem: docker
+  directory: "/server"
+  schedule:
+    interval: daily
+    time: "11:00"
+  open-pull-requests-limit: 5
+
+# cloudbuild.yaml
+ - package-ecosystem: "docker"
+  directory: "/"
+  schedule:
+    interval: daily
+    time: "11:00"
+
+# requirements*.txt
+- package-ecosystem: pip
+  directory: "/"
+  schedule:
+    interval: daily
+    time: "11:00"
+  open-pull-requests-limit: 5
+ 
+
+
+
+
+
+
+  
+
+
+
+
+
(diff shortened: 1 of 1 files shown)


AlexCLeduc:
Thanks Lila! 

Daily sounds annoying but I trust you guys on this 😄 

I don't think assigning the PRs to someone is necessary, although I am worried about people merging them without testing them. 

I agree dependabot should be added to the helpers repo. We do need to revise the responsibilities a bit there because we're relying a little too heavily on Luc. I don't even know if it auto-releases or if he manually creates pypi releases. I think he's on vacation for a couple weeks so we can revisit later

AlexCLeduc (approved):

lucbelliveau:
> Thanks Lila!
> 
> Daily sounds annoying but I trust you guys on this 😄
> 
> I don't think assigning the PRs to someone is necessary, although I am worried about people merging them without testing them.
> 
> I agree dependabot should be added to the helpers repo. We do need to revise the responsibilities a bit there because we're relying a little too heavily on Luc. I don't even know if it auto-releases or if he manually creates pypi releases. I think he's on vacation for a couple weeks so we can revisit later

The helpers repo does automatic deployments to PyPi FYI.

LilaKelland:
Great! Luc can we assume it's okay then to let Dependabot to merge minor and patch updates?

AlexCLeduc:
I would caution against letting it auto-merge anything, I've seen python packages break from minor updates and the helpers' does not have high test coverage

PR #66, opened 2023-07-17
Pattern for more safely using sensitive/secret config values in dev

Stephen-ONeil:
As discussed in #50, there's good reason to commit a dev `.env`  to the repo (one less thing for each individual dev to manage/keep in sync, and needs to be available and in-sync for testing in CI too), but there's also a risk that dev's may at some point need to use a sensitive/secret env value for local testing. Committing the dev `.env` file makes it much more likely that such a value could be inadvertently committed in the future, which we do not want.

GitHub secret scanning and push protection are an incomplete mitigation, as they won't be able to detect all possible secret patterns. Better than nothing though.

I'm introducing a further convention here, where a git-ignored `.env.dev-secret` can be used if a dev ever needs to test with a secret value locally (e.g. an API key for a third party service, although even then it should be a temporary, non-prod, key). If present, the values from `.env.dev-secret` will be merged in (with priority over) `.env.dev-public` when accessed in `settings.py`.

This convention existing, and renaming  `.env.dev` to `.env.dev-public`, should make it even less likely for devs to slip up and leak a secret.

### DIFF
--- a/server/server/settings_util.py
+++ b/server/server/settings_util.py
@@ -1,27 +1,37 @@
 import os
 
-from decouple import Config, RepositoryEnv, undefined
+from decouple import Config, RepositoryEnv, UndefinedValueError, undefined
 
 PROD_ENV_FILE_NAME = ".env.prod"
 DEV_PUBLIC_ENV_FILE_NAME = ".env.dev-public"
 DEV_SECRET_ENV_FILE_NAME = ".env.dev-secret"
 
 
+# Reminder: decouple config(...) looks in OS env vars first, configured env file second
 def get_project_config(BASE_DIR):
     try:
+        # If the prod env file exists, use it exclusively
         config = Config(
             RepositoryEnv(os.path.join(BASE_DIR, PROD_ENV_FILE_NAME))
         )
-    except:
+    except FileNotFoundError:
+        # If the prod env file doesn't exist
+        #   1) look for the secret in the dev-secret env file
+        #   2) if the dev-secret env file doesn't exist OR the the secret isn't
+        #      found in it, look in the dev-public file
+        #   3) if still not found, fall back to provided `default` (if any)
 
+        # IMPORTANT: dev_config_merged must support the API of decouple's config function! Masquerades as it in use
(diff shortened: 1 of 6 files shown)


server/server/settings_util.py
```diff
@@ -0,0 +1,33 @@
...
+            except:
+                return Config(
+                    RepositoryEnv(
+                        os.path.join(BASE_DIR, DEV_PUBLIC_ENV_FILE_NAME)
+                    )
+                )(*args, default, **kwargs)
+
+        config = dev_config_merged
+
+    return config
```
AlexCLeduc, line 33: I trust the tests but I'm super confused, can configs be lazy? How come this function can return a `Config` instance OR a function that resolves to a `Config` instance? Can we add a comment?

Where is the merge happening? It seems we only use the public file if the secret one doesn't exist. Ideally I'd like to put values like `ALLOWED_HOSTS=*` in the public-dev and _only_ things like `GC_NOTIFY_KEY` in the secret-dev but it looks like I'd have to repeat everything from the public-dev in the secret-dev. Unless decouple magic is throwing a special exception when a value lookup fails? If so, can we use that exception subclass and add a corresponding comment?

server/tests/test_settings_util.py
```diff
@@ -0,0 +1,73 @@
+import pytest
+from decouple import UndefinedValueError
```
AlexCLeduc, line 2: Ahh, this answers my above question 😄

AlexCLeduc (changes requested):
Thanks, this is great! The pattern will come in handy for our DEEPL/GC_NOTIFY keys

This is likely to into the template and/or the helpers repo at some point, so please document the weird decouple API stuff

Can we also update the README to use `.env.dev-public` rather than `.env` ?

AlexCLeduc (approved):
Ah I hadn't noticed the .env is in the repo now, a bit strange compared to our other projects. Some people like to disable the toolbar, might be annoying to have have those types of changes constantly tracked but this is probably easier for new devs and forcing awareness around the toolbar is a bonus, people will finally notice N+1 issues 😄

PR #69, opened 2023-07-17
Logging, with structured logging and (optional) slack alerts for prod errors

Stephen-ONeil:
Added a non-default logging configuration, with some extra steps to guarantee consistency between logs made via either the standard python `logging` module or the `structlog` package. We should probably standardize on one or the other, but the consistency is good for log sources we don't control (third part packages, etc.). 

Most usefully, all logging in prod now uses JSON formatting, so processing/exploring them will be far more   

Also added `django_structlog` middleware for more granular request logging, and to associate user ID, IP address, etc to logs where ever possible. We'll want to double check whether we want to be logging user IPs, from a privacy perspective. If it comes down to it, we could trim and/or hash them perhaps.  

Example of "request started" log from `django_structlog.middlewares.RequestMiddleware` (logged via `structlog`)
image

Example of AXES event log (logged via standard `logging` module)
image

My log handler for sending prod error alerts to slack is in there too, but I've not configured it.

### DIFF
--- a/server/.env.dev-public
+++ b/server/.env.dev-public
@@ -28,3 +28,6 @@ FORCE_WHITENOISE_PROD_BEHAVIOUR=False
 # this is to disable session timeout
 PHAC_ASPC_SESSION_COOKIE_AGE=99999999 # this doesn't seem to work?
 PHAC_ASPC_SESSION_COOKIE_SECURE=0
+
+# set to DEBUG if you want to get hit with the firehose
+LOWEST_LOG_LEVEL="INFO"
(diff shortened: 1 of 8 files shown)


Stephen-ONeil:
Writing tests for the logging configuration itself might be a TODO, although what to test is a bit tricky here. Maybe assert my claim that logs will be consistent between `structlog` and `logging`? Hm.

AlexCLeduc (approved):
This is really neat, thanks! Are the screenshots from the GCP console?

I think we will have to scrub IPs (or part of them?) eventually once the API goes public. Might have to configure axes at that point too if it isn't already

Stephen-ONeil:
GCP console, yup.

Issue #57, opened 2023-07-07
System check warning re AxesModelBackend was renamed


LilaKelland:
Get this warning ```?: (axes.W003) You do not have 'axes.backends.AxesStandaloneBackend' or a subclass in your settings.AUTHENTICATION_BACKENDS.
	HINT: AxesModelBackend was renamed to AxesStandaloneBackend in django-axes version 5.0.when running ```on each step when running 
```
python ./manage.py loaddata cpho/fixtures/periods.yaml, 
python ./manage.py loaddata cpho/fixtures/dimension_lookups.yaml
python ./manage.py runscript cpho.scripts.dev
```

LilaKelland:
adding "axes.backends.AxesStandaloneBackend" to AUTHENTICATION_BACKENDS in settings.py fixes this

PR #73, opened 2023-07-20
Testing as a Cloud Build step and code coverage reporting.

LilaKelland:
Used docker compose to be able to run tests as a cloud build step (some tests seemed to need to a database running).  

image

Added .coveragerc file in /server.  Are there other files/ patterns we should be ignoring?
   ```
tests/**
    **/__init__.py
    **/migrations/*
    **/monkey_patch_for_testing.py  
```

Items we'll need to address in the future:
-  In order to run the tests, the Dockerfile has the dev and formatting requirements installed as well - we'll need to split them out for the final build image as Stephen suggested. 
- Have the coverage report printed (more visibly) back to the repo, rather than deep in the cloudbuild logs. 
- Have build fail if test coverage is below a particular threshold

### CODE REVIEW

File: server/Dockerfile [hunk 1]
```diff
@@ -25,13 +25,15 @@ RUN pip install --upgrade pip
 
 # install dependencies
 COPY ./requirements.txt .
-RUN pip wheel --no-cache-dir --no-deps --wheel-dir /usr/src/app/wheels -r requirements.txt
+RUN pip wheel --no-cache-dir --no-deps --wheel-dir /usr/src/app/wheels -r requirements.txt  
 
+# TODO: this is a workaround, to be replaced when working out a proper data migration approach, possibly
+# when using entrypoint.prod.sh to seed dat, need dev requirements installed. Also needed for testing.
 # TODO: this is a workaround, to be replaced when working out a proper data migration approach, possibly
 # when using entrypoint.prod.sh to seed dat, need dev requirements installed. Otherwise, leave this commented out
```
Stephen-ONeil, line 33: Comment duplicated?

server/.coveragerc
```diff
@@ -0,0 +1,6 @@
+[run]
+omit =
+    tests/**
+    **/__init__.py
+    **/migrations/*
+    **/monkey_patch_for_testing.py
```
Stephen-ONeil, line 6: Maybe ignore `./manage.py`, `server/wsgi.py`, and `server/asgi.py`, but those are very minor. I've seen some other django projects ignore `*/scripts/*`, but I think that's arguable.


This should be a follow up PR/issue for later (unless you look in to it and it is trivial to set up), but looking at other project's `.coveragerc` files reminded me that there's a coverage plugin called `django_coverage_plugin` we can add to possibly get coverage reports on django templates as well. Note, the example I saw this in was using django's built in templates, while this project is using jinja for templating. Haven't checked if this specific plugin can also track coverage for jinja files, or if some jinja alternative exists.
```
plugins =
    # allows coverage for templates
    django_coverage_plugin
```

Stephen-ONeil (changes requested):
Still looking at the cloud build and the docker compose components, and will answer your specific questions from the PR comment too. This is a partial review for the other files just to move the review process along in the meantime :+1:

deploy/gcloud_init_setup.sh
```diff
@@ -83,7 +83,6 @@ if [[ "${build_skip}" != "S" ]]; then
       --include-logs-with-status \
       --no-require-approval
   fi
-fi
```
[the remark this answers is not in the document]
LilaKelland: ooh - nice catch - thank you!  I had pulled that out accidentally when I stripping out codecov.

cloudbuild.yaml
```diff
@@ -19,7 +19,50 @@ steps:
...
+      #!/usr/bin/env bash
+      set -o errexit
+      set -o pipefail
+      set -o nounset
+      
+      echo "Get cloud build project configuration values"
+      source .env.cloud_build
+
+      echo "Build new server and postgres db images for testing"
+      docker compose --env-file .env.cloud_build up --build -d
```
Stephen-ONeil, line 34: I see `up` is used here to start the container used in `"Run tests and produce coverage report"` and eventually killed in `"Tear down from tests"`. I'd prefer the `up` and `down` to happen within the same Cloud Build step (meaning along with running the tests, although building the composed image could still be separate probably). Let me know if there was some particular reason to split things in to multiple steps!

The reason I want the life time of the container to be contained to one step is because I expect, as some point, we'll start running different cloud build steps in parallel. I think things will get confusing/possibly buggy at that point if steps aren't self contained. Maybe that's not the right way to think about Cloud Build steps though, I'm still not fully comfortable with them in some ways.

docker-compose.yaml
```diff
@@ -0,0 +1,47 @@
...
+    image: ${IMAGE_NAME_FOR_RUN}
+    command: >
+      sh -c "
+      python ./manage.py migrate && \
+      python ./manage.py loaddata cpho/fixtures/dimension_lookups.yaml && \
+      python ./manage.py loaddata cpho/fixtures/periods.yaml && \
+      python ./manage.py runscript cpho.scripts.dev && \
+      exec gunicorn --bind 0.0.0.0:8080 --workers 1 --threads 8 --timeout 0 server.wsgi:application"
+    ports:
+      - 8080:8080
```
Stephen-ONeil, line 29: Where these `command` steps necessary for running the tests? Or was this for testing the image locally? Should be able to remove it all, including the exposed port, and still run tests in this image. Maybe even give an empty `command` value to overwrite the image default, or replace it with the command to execute the tests

Stephen-ONeil: I guess I'm assuming that this `docker-compose.yaml` is intended to be single purpose for running tests in cloud build. If it is, drop these commands and maybe rename it something like `docker-compose.cloud-build-test-env.yaml`.

docker-compose.yaml
```diff
@@ -0,0 +1,47 @@
...
+    env_file:
+      - ./server/.env.dev
+    environment:
+      - DB_HOST=db
+      - POSTGRES_HOST_AUTH_METHOD=trust
```
Stephen-ONeil, line 34: Might be worth adding a comment here to note that values set in `environment` will effectively over-write the values in `.env.dev`, because `decouple.Config` looks in OS env vars before .env files. Important because if that behaviour ever changes, or if code running in this container reads from the `.env` file without using `decouple`, then this won't work for setting the `DB_HOST` value any more.

server/Dockerfile
LilaKelland: Do I put commit tags in here? Realizing my commit comments need to be more descriptive: [removed douplicate comment per Stephen's review](https://github.com/PHACDataHub/cpho-phase2/pull/73/commits/189b4e19d4725e0303b6fe3595d7af9b06779572) [189b4e1](https://github.com/PHACDataHub/cpho-phase2/pull/73/commits/189b4e19d4725e0303b6fe3595d7af9b06779572)

cloudbuild.yaml
```diff
@@ -19,7 +19,50 @@ steps:
...
+      set -o errexit
+      set -o pipefail
+      set -o nounset
+
+      echo "Get cloud build project configuration values"
+      source .env.cloud_build
+
+      echo "Run tests and create coverage report"
+      docker compose --env-file .env.cloud_build exec -T server coverage run -m pytest
+      docker compose --env-file .env.cloud_build exec -T server coverage report --show-missing
```
Stephen-ONeil, line 49: Probably worth adding a TODO comment here noting that uploading the coverage report somewhere is a next-step, and that for now coverage reports need to be dug up in the Cloud Build logs. Would be a helpful breadcrumb for anyone looking for the coverage reports in the meantime, and a good reminder (in addition to the issue about it).

Stephen-ONeil (changes requested):
I have thoughts on how to rearrange the docker file/build layers to keep the dev-requirements out of the prod image, but I'm happy to leave that as a follow up PR so that we can bring CI testing to the repo sooner :+1:

server/Dockerfile
Stephen-ONeil: Ooo, I'd appreciate if you do! Usually, either when I see commits to a branch I'm reviewer on or when prompted for re-review, I just look for comments that GitHub has marked "outdated" (meaning some line of code in the comment's scope has since changed) and try to figure out my self whether it was resolved or not, haha.

I always consider it the reviewer's responsibility to follow up and hit the resolve button, but I don't know if everyone operates that way.

cloudbuild.yaml
LilaKelland: Yep - this makes sense!  Had them split out before as they were sandwiching the "upload to codecov step"(that also downloaded the uploader).  Will make this change.

.gitignore
```diff
@@ -25,3 +25,5 @@ runserver.sh
 **/logfile
 
 server/logfile
+
+.coverage
```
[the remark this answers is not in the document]
LilaKelland: [e37c5b7 added ** in front of coverage so to ignored anywhere in tree](https://github.com/PHACDataHub/cpho-phase2/commit/e37c5b7dcba3ce33c0b8df2a017b66128de766bb)

cloudbuild.yaml
LilaKelland: Tests (including build and tear down have been consolidated into one step).

[Removed extra steps in build - now the build images, run tests and te…](https://github.com/PHACDataHub/cpho-phase2/commit/936248c9616e6353f76c34a11dc60f95b7d4f7df)

[Simplified title for Run tests step of cloudbuild.](https://github.com/PHACDataHub/cpho-phase2/commit/179ede0e5af3835f74e01584ccccddea31ce05a1)

docker-compose.yaml
LilaKelland: [Changed network flag to external: true to be able to work in cloud bu…](https://github.com/PHACDataHub/cpho-phase2/commit/f671166efc11b7d2548eac3ec27e6e1b418bb455) 

Was unsure if we should simplify the run line exec gunicorn --bind 0.0.0.0:8080 --workers 1 --threads 8 --timeout 0 server.wsgi:application  to something like exec gunicorn  server.wsgi:application as other flags not needed here, or to keep it as close to run command as possible?

Stephen-ONeil: I might be missing something, but I meant that if all this container does is run tests then `command` could just be `coverage run -m pytest` and ports could be closed. The Django test runner handles any migrations if necessary and `server/tests/conftest.py` loads the necessary fixture data. `cpho.scripts.dev` and starting gunicorn isn't necessary for running the tests either.

cloudbuild.yaml
LilaKelland: Issue #75  & [Added a TODO to upload coverage report as a next step and make report…](https://github.com/PHACDataHub/cpho-phase2/commit/ec97c7e6505b59165bb5ae66e8f9a7ce8f4af5ce)

docker-compose.yaml
LilaKelland: You're absolutely right - iteratively getting there!

LilaKelland:
Added  Issue #78 Rearrange the docker file/build layers to keep the dev-requirements out of the prod image

server/dev-db-init.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+set -o errexit
+set -o pipefail
+set -o nounset
+
+# Get environment variables
+source $(dirname "${BASH_SOURCE[0]}")/.env.dev
+POSTGRES_USER=postgres
```
Stephen-ONeil, line 8: Minor bash convention comment, capital variable names are usually for environment variables. Local variables (meaning they aren't turned in to env vars with `export <variable name>`) should use lower case names :+1:

notes.txt
Stephen-ONeil, line 1: Guessing this wasn't meant to be committed?

docker-compose.run-tests.yaml
```diff
@@ -0,0 +1,48 @@
...
+      - ./server/.env.dev:/docker-entrypoint-initdb.d/.env.dev
+      - ./server/dev-db-init.sh:/docker-entrypoint-initdb.d/dev-db-init.sh
+    environment:
+      - POSTGRES_HOST_AUTH_METHOD=trust  
+    env_file:
+      - ./server/.env.dev
+
+  server:
+    build: ./server
+    image: ${IMAGE_NAME_FOR_RUN}
```
Stephen-ONeil, line 21: Is image name needed if building from a docker file? Since the `${IMAGE_NAME_FOR_RUN}` image is actually built in a later step, and only on main, does naming the image built at this step potentially conflict with that?

Alternatively, if the "build image if main" step was moved to before the test step, and done on all branches, could the docker compose just reference that image instead of building here? Just curious, not suggesting you make that change because I doubt it will work that way post-#78.

Stephen-ONeil (changes requested):
Just some minor comments and a question, almost appoved!

server/server/settings.py
```diff
@@ -222,6 +222,7 @@
 # https://docs.djangoproject.com/en/4.1/ref/settings/#auth-password-validators
 
 AUTHENTICATION_BACKENDS = [
+    "axes.backends.AxesStandaloneBackend",
```
Stephen-ONeil, line 225: See #80. I looked at this and issue #57, and my first thought was that it's something the `phac_helpers` library should be enforcing, since that's where `AxesMiddleware` is added and configured from. My second thought was to check if it already does, and it turns out the answer is yes, as long as the project calls `configure_authentication_backends` haha

LilaKelland: This was failing (can't remember which build step) without adding this, but there was an error indicating that it needed this, and when adding configure_authentication_backends in, it seemed to resolve the issue (so I added it).  But this is when everything was failing.  

Just removed this line, and tried it again - and you're right it works!  There's just a warning - so keeping this out. 
image

server/dev-db-init.sh
LilaKelland: Great to know!  I've removed the variable completely and just added postgres inline.  (Or is it better practice to keep this as a variable?)

notes.txt
LilaKelland: Ugh - this is why I try do less git add -A.  Yes - mistake - thank you!

docker-compose.run-tests.yaml
LilaKelland: IMAGE_NAME_FOR_RUN is assigned in make_cloud_build_env_file.sh in the first Cloud Build step, and works because we pass that variable in with --env-file .env.cloud_build in the docker compose up line (which it appears we don't need anymore - because the image name is irrelevant as this is only used for contained testing).  I've removed those flags from the testing cloud build step and hard coded the image name in docker-compose.run-tests.yaml. May need to revisit in the next step to be able to access SHA to tag reports when saving.

PR #79, opened 2023-07-24
Simplify implementation of get_project_config

Stephen-ONeil:
Realized there actually was a first-party way to merge env files with `decouple`


Stephen-ONeil:
Trivial and passing tests, merging

Issue #84, opened 2023-07-25
Add step to enable Postgres audit logs in gcloud_init_setup.sh


Stephen-ONeil:
As described [here](https://cloud.google.com/sql/docs/postgres/pg-audit#enable-auditing-flag). Setting/updating database flags is easy, but you also need to run `CREATE EXTENSION pgaudit;` in the database. So we're back to needing ~a good script-able way to connect to the prod DB or~ a good strategy for DB script Cloud Run jobs.

~Could be~ Cloud SQL Auth Proxy, but that's an additional dependency needed on-machine and might be difficult to integrate smoothly in the script. It also can't currently connect to the DB; either the dev machine needs to be able to connect to the VPC to use the private IP, or we need to be temporarily enabling a public IP before connecting via Cloud SQL Proxy. Hm. See #64, 

Much less hacky would be to make a one-time DB initialization Cloud Job that we kick off from the init script, although I know there's still a desire to have a way to get a hands-on DB shell, at least during this pre-prod stage.

... see also [these limitations and warnings](https://cloud.google.com/sql/docs/postgres/pg-audit#limitations). Could need some caution and careful configuration (especially if we want to use this on a busier application in the future).

Stephen-ONeil:
We can also add [enabling Query Insights](https://cloud.google.com/sql/docs/postgres/using-query-insights) to our initialization script. It's limited to only one week of data, but it's got an easier cloud dashboard for quickly looking at longest running queries etc.

Stephen-ONeil:
From #85, we should have OpenTelemetry tracing comments added to our SQL queries already, they should end up in the audit logs and should (?) be picked up and associated to the Cloud Trace explorer from there.

> With `DjangoInstrumentor`'s `is_sql_commentor_enabled=True` configuration, "tracing comments" will be appended to all of our ORM SQL calls. This is just a bunch of metadata in a comment, including the trace ID, that's sent to the DB _in_ the query, and then ends up in the DB logs for that query as a consequence. Weird hack, but the `is_sql_commentor_enabled` flag is based on a SQL tracing library that Google wrote and donated to the OpenTelemetry project, so it should work
 > We might want to disable some of the default data crammed in there, see [the docs](https://opentelemetry-python-contrib.readthedocs.io/en/latest/instrumentation/django/django.html#sqlcommenter-configurations)
[...]

PR #85, opened 2023-07-26
OpenTelemetry Request Tracing

### LINKED ISSUE #23: CPHO: Add OpenTelemetry to server
OpenTelemetry is an open observability standard. [Google's Cloud Trace works with OT](https://cloud.google.com/trace/docs/setup/python-ot) and there is [a library](https://opentelemetry-python.readthedocs.io/en/latest/examples/django/README.html) that automatically instruments Django applications. We should get this working on the CPHO server.

Stephen-ONeil:
Closes #23 (WIP)

(sorry reviewers, LOTS of new concepts and weirdness to this one)

OpenTelemetry is an ecosystem for three types of observability:
  1) Request tracing
  2) Metrics and metric aggregation
  3) Logs
  
This PR sets up our request tracing instrumentation. Metrics will likely come later. We will likely not use their logging features any time soon, as their log specification is still WIP. Interfacing with GCP Cloud Logging directly will be simpler and more stable for us anyway.

Request tracing is all about following a request as it moves through our network, collecting timing information and metadata at each step. With the right configuration, it can also associate all of the related logs each piece of infrastructure makes while handling the request, and give us a single, convenient place to view all the data we have on a given request's lifecycle. Each request has one "trace". "Spans" represent discrete stages of the request's handling, and are hierarchical. For us, a span is most often going to be 1-1 with a piece of infrastructure in our network, and a span's start and end times will be the times when a request is received and a response sent, respectively.

See the"Trace Overview" and "Trace Explorer" pages in the project's GCP Console for an idea of what all this means in practice. You may want to visit the live site first to generate some fresh traces/logs. 

Note:
  - the `X-Cloud-Trace-Context` header is GCP's standard way to propagate tracing context for each request
  - `X-Cloud-Trace-Context` [is automatically added to inbound requests for Cloud Run](https://cloud.google.com/trace/docs/overview#configurations_with_automatic_tracing), for automatic latency data collection. This is something we need to account for. We'll want to use and propagate this trace context BUT, as detailed in [this blog post](https://anecdotes.dev/opentelemetry-on-google-cloud-unraveling-the-mystery-f61f044c18be), the GCP behaviour is to only sample _some_ requests for the automatic latency data (actually, seems that instead of picking regular traffic to sample, it makes dedicated sampled ping requests to `/`). The way this works, every request to Cloud Run gets a `X-Cloud-Trace-Context` header, but most have a flag indicating them as "non-sampling". That blog post lists some options for how to deal with this if you want full tracing (we likely do, given this isn't a super high traffic application).
    -  resolution: `sampler=sampling.ALWAYS_ON` used in the app trace provider, along with `CloudTraceFormatPropagator`. This results in everything being sampled starting once it reaches the application, but (almost) all traces include a missing span for the load balancer step (because the trace context and sample ID from the load balancer always exist and propagate to the application, but their spans aren't always sampled/exported back to Cloud Trace). Something to address in or after #76, when we switch off the default Cloud Run load balancer

TODO:
  - [x] Application instrumentation (start [here](https://cloud.google.com/trace/docs/setup/python-ot))
    - [x] simple dev-local OpenTelemetry tracing with `ConsoleSpanExporter`
      - [x] add env var for enabling/disabling this, will probably want it disabled by default in dev as it really clutters the console output
    - [x] configure OpenTelemetry trace exporting for GCP
      - [x] use [`opentelemetry-exporter-gcp-trace`](https://pypi.org/project/opentelemetry-exporter-gcp-trace/)
        - this lets us export directly to Cloud Trace; means we don't need the Collector sidecar container setup (yet) 
      - [x] use [`opentelemetry-propagator-gcp`](https://pypi.org/project/opentelemetry-propagator-gcp/)
    - [x] as per [these](https://cloud.google.com/trace/docs/trace-log-integration#associating) docs, add `trace` and `spanId` fields to app logs associated with a given request, to connect them together
      - we can do with with `structlog.contextvars.bind_contextvars`. Additional `structlog` context vars end up in both `structlog` and standard `logging` logs with our logging config, as asserted in `test_json_logging_consistent_between_standard_logger_and_structlogger`
    - [x] add resource detection with [`opentelemetry-resourcedetector-gcp`](https://pypi.org/project/opentelemetry-resourcedetector-gcp/) ([e.g.](https://google-cloud-opentelemetry.readthedocs.io/en/latest/examples/cloud_resource_detector/README.html))
  - [x] add reminders about integrating tracing on upcoming infrastructure PRs/issues
     - with `opentelemetry-propagator-gcp`, the automatic Cloud Run load balancer is showing up in our traces (most of the time, I think it only samples for tracing sometimes). Might turn out that most managed services integrate with Cloud Trace automatically
      - #76
      - #84
   - [x] Open issues for these next-steps:
      - metrics (Opened as #97)
        - use [`opentelemetry-exporter-gcp-monitoring`](https://pypi.org/project/opentelemetry-exporter-gcp-monitoring/) for monitoring metrics
        - this is likely when we'll want a Cloud Run sidecar to run an OpenTelemetry [Collector](https://opentelemetry.io/docs/collector/#when-to-use-a-collector), assuming we want metrics with aggregation logic, which is the super power of OpenTelemetry metrics. See the google docs [here](https://cloud.google.com/run/docs/tutorials/custom-metrics-opentelemetry-sidecar) and [here](https://cloud.google.com/run/docs/deploying#sidecars)
          - TBC do the predefined aggregation types require a collector, or can the GCP Cloud Trace backend perform them?
      - DB tracing/log association (Added to #84)
        - with `DjangoInstrumentor`'s `is_sql_commentor_enabled=True` configuration, "tracing comments" will be appended to all of our ORM SQL calls. This is just a bunch of metadata in a comment, including the trace ID, that's sent to the DB _in_ the query, and then ends up in the DB logs for that query as a consequence. Weird hack, but the `is_sql_commentor_enabled` flag is based on a SQL tracing library that Google wrote and donated to the OpenTelemetry project, so it should work
        - we might want to disable some of the default data crammed in there, see [the docs](https://opentelemetry-python-contrib.readthedocs.io/en/latest/instrumentation/django/django.html#sqlcommenter-configurations)
        - confirm that this actually gets them associated to the intended traces. The Cloud Trace docs mention that it only works for structured logging (JSON payload logs), but unless there's some magic somewhere the added query comments seem like they would just wind up in text payload logs?


server/tests/test_config_util.py
```diff
@@ -20,12 +24,38 @@ def write_env_file(temp_path, env_file_name):
...
+
+
+def test_is_running_tests_returns_false_outside_test_execution_environment():
+    # hacky, but has to be hacky for a test to assert something that by definition _can't_
+    # occur while in the test execution environment
+
+    manage_py_path = os.path.join(
+        os.path.dirname(os.path.dirname(os.path.abspath(__file__))),
+        "manage.py",
+    )
```
AlexCLeduc, line 38: This is nice to have but just FYI you don't have to be this thorough 😄

server/server/open_telemetry_util.py
```diff
@@ -0,0 +1,112 @@
...
+                else open(os.devnull, "w")
+            )
+        )
+
+        resource = ProcessResourceDetector(raise_on_error=True).detect()
+    else:
+        project_id = requests.get(
+            "http://metadata.google.internal/computeMetadata/v1/project/project-id",
+            headers={"Metadata-Flavor": "Google"},
+        ).text
```
AlexCLeduc, line 50: We have to make an HTTP request just to figure out the project ID? That just seems really strange, isn't it usually also available as an environment variable?

Stephen-ONeil: Not in Cloud Run, by default. Agree that it's weird, I know I had some reason in mind for going with this, maybe just because it's the "official" way to do things? Could look for an env var first and fall back on this.

AlexCLeduc (approved):
Haven't played around with it yet but seems pretty cool and doesn't seem to get in our way at all, appreciate the small footprint in settings/application code. Would love to see a demo in GCP at some point

Stephen-ONeil:
> Would love to see a demo in GCP at some point

Happy to give a demo, maybe at or after the touch point next week?

LilaKelland (commented):
Looks great! Curious for my own understanding - why wsgi over asgi?

AlexCLeduc:
We won't be able to use asgi for a long time, we'd have to replace all our DB/ORM calls with async equivalents, which would take a little while for first-party code but much longer for 3rd party tools. 

BTW, the graphQL configuration I will setup soon will rely on a custom promise implementation, which doesn't play well with async. 

It's best to assume we won't be using async in any django projects for at least the next couple years.

LilaKelland:
Makes sense - thanks Alex!

PR #89, opened 2023-07-26
UI improvements, (breadcrumb, period selection, )

AlexCLeduc:


server/tests/test_manual_data_input.py
```diff
@@ -279,3 +273,61 @@ def test_modify_all_dimensions(vanilla_user_client):
         ).value
         == 7.5
     )
+
+    # check periods correctly assigned on all records
+    assert not ind.data.exclude(period=period)
+
+
+def test_non_changes_dont_create_versions(vanilla_user_client):
```
AlexCLeduc, line 281: Decided to move these to their own tests. Although we had something like this before (see deleted code) it actually wasn't testing stuff properly because it the "previous" records were not assigned the proper period

server/cpho/jinja2/review_data.jinja2
```diff
@@ -1,6 +1,14 @@
...
 {% from 'generic_macros.jinja2' import date_display, todo_block %}
 {% from 'cpho_macros.jinja2' import indicator_data_table_w_submissions, submission_status_badge %}
+{% import 'breadcrumb_macros.jinja2' as bc %}
+
+{% block content_breadcrumbs %}
+  {{ bc.indicators_index() }}
+  {{ bc.indicator(view.indicator) }}
+  {{ bc.indicator_period(view.indicator, period) }}
+  {{ bc.item(tdt("Review for submission") , None, True) }}
+{% endblock %}
```
AlexCLeduc, line 11: This is how we can use breadcrumbs in a semi-D.R.Y way without leaking presentational concern in the views

server/cpho/urls.py
```diff
@@ -22,8 +22,8 @@
     ),
     path(
         "indicators/<int:pk>/<int:period_pk>/",
-        views.ViewIndicatorForYear.as_view(),
-        name="view_indicator_for_year",
+        views.ViewIndicatorForPeriod.as_view(),
+        name="view_indicator_for_period",
```
AlexCLeduc, line 26: This change is responsible for a large number of diffs

server/cpho/jinja2/indicators/view_indicator.jinja2
```diff
@@ -17,17 +25,32 @@
...
+        {% for period,data_count in data_counts_by_period.items() %}
+          {% if data_count == 0 %}
+            <li>
+              <a href="{{ url('view_indicator_for_period', args=[indicator.id, period.pk]) }}">{{ tdt("Add data for ") }} {{ period }}</a>
+            </li>
+          {% endif %}
+        {% endfor %}
+      </ul>
+    </div>
+  </div>
```
AlexCLeduc, line 54: I split up the period selection, 

Screenshot 2023-07-26 at 17 57 03

server/cpho/views/indicator_data.py
```diff
@@ -232,6 +232,7 @@ def age_group_formset(self):
         for form in fs:
             # new unsaved instances' save() crash when no dimension type is specified
             form.instance.dimension_type = age_dimension
+            form.instance.period = self.period
```
AlexCLeduc, line 236: There was previously a bug where you would add new age group data, but not see it show up. It was in fact creating records, but since they weren't linked to any period they would not appear. We may want to add a non-null constraint on the period field of indicator-datum

PR #92, opened 2023-07-26
Added rules framework and rules

msarar:
initial rules framework inplemented
we have 3 users now: admin, hso and program


msarar:
I see that the tests are failing now and im not surprised.
all the failures are Suspicious operation errors for 'vanilla user'
Need to fix tests before merging

AlexCLeduc:
Rather than use different types of users, use `patch_rules` with relevant rules to fix tests

server/cpho/rules.py
```diff
@@ -0,0 +1,105 @@
...
+        is_admin(user)
+        or is_hso(user)
+        or (is_program(user) and is_indicator_lead(user, indicator))
+        or (
+            is_program(user)
+            and is_branch_lead(user)
+            and indicator_of_users_branch(user, indicator)
+        )
+    )
+
```
AlexCLeduc, line 56: Rules can be combined using bool operators, this can be re-written as 

```python
rules.add_rule("can_edit_indicator", (
  is_admin | 
  is_hso | 
  (is_program & is_indicator_lead) | 
  ( is_program & is_branch_lead & indicator_of_users_branch )
)
```
It's a question of style, though, it won't always work if you need to pass a different variable, but it should work in this case because the rules library will ensure to only pass the indicator argument to rules who take 2 arguments

server/cpho/views/indicators.py
```diff
@@ -78,6 +81,16 @@ class ViewIndicator(TemplateView):
...
+    def get(self, request, *args, **kwargs):
+        context = self.get_context_data(**kwargs)
+        if not test_rule(
+            "can_edit_indicator", request.user, context["object"]
+        ):
+            raise SuspiciousOperation(
+                f"User {request.user} cannot edit this indicator"
+            )
+        return self.render_to_response(context)
+
```
AlexCLeduc, line 93: I think we'll instead want 1 or 2 "base" mixins and potentially even more if you find yourself repeating yourself often,

The more standard way of doing auth checks in CBVs is with the dispatch method, because it's defined on the base view class and is fired immediately before any other methods when a request is sent. It gets run regardless of GET/POST/PUT/etc.  

```python
class MustPassAuthCheckMixin(View):
    def check_rule(self):
        raise NotImplementedError("must override check_rule") 

    def dispatch(...)
          if not self.check_rule():
               raise PermissionDenied()

          return super().dispatch(..)          
          
```
Many rules don't take arguments, so you can inherit this with a shorthand for those rules,
```python
class MustPassUserRuleMixin(MustPassAuthCheckMixin):
    def check_rule(self):
             if not getattr(self,rule_name,None):
                 raise TypeError("must define rule_name on class"

              return test_rule(sefl.rule_name, self.request.user)
```

In the cases you want a view to check one rule for GET and a different one for POST, it might be easier to just code that logic in the view itself, again usually in the dispatch. Or you can write a mixin with 2 different check methods 🤷

server/cpho/views/submission.py
```diff
@@ -113,3 +134,25 @@ def get_context_data(self, *args, **kwargs):
...
+                    "can_submit_as_program", request.user, self.indicator
+                )
+            )
+            or (
+                test_rule("is_hso", request.user)
+                and test_rule(
+                    "can_submit_as_hso", request.user, self.indicator
+                )
+            )
+        ):
```
AlexCLeduc, line 154: Views combining several rules is an anti-pattern, usually it means this expression should be its own rule

server/cpho/views/submission.py
```diff
@@ -113,3 +134,25 @@ def get_context_data(self, *args, **kwargs):
...
+            raise SuspiciousOperation(
```
AlexCLeduc, line 155: SuspiciousOperation returns a 400 (Bad request) vs PermissionDenied returns 403 (Unauthorized). I don't think anyone in security will care but I think this is more of a 403 than a 400 🤷 

A 400 would be more appropriate if a user bypassed frontend validation like a readonly field or something, we don't use it nearly as often as PermissionDenied

server/cpho/rules.py
msarar: After discussion with Alex, decided to keep this as the syntax is probably easier to understand

AlexCLeduc (approved):

Issue #95, opened 2023-07-28
Fail main branch build if test coverage is below a given percentage.


LilaKelland:
Building on the outcome of PR #93.  Probably need to look at branch thresholds/ and compare to last build too - will look at other projects and see solutions to see if can borrow something similar.

LilaKelland:
Closed with PR #96

Issue #75, opened 2023-07-21
Upload coverage reports & make more visable.


LilaKelland:
This issue comes out of PR #73 Testing as a Cloud Build step and code coverage reporting. 

The test coverage report is currently printed deep in the cloud build logs.  Let's pull this out and ideally have it printed back to GitHub, but saved to Google Cloud Storage is a good first step. 

[Link to comment.](https://github.com/PHACDataHub/cpho-phase2/pull/73#discussion_r1269927150)

PR #96, opened 2023-07-28
Fail build below coverage threshold

LilaKelland:
This will close issue #95 and builds on PR #93.  Thinking that maybe PR #93 and this one could be combined? 

We have test coverage reporting at this point, but want to be able to use it.  

(From #93 Will close https://github.com/PHACDataHub/cpho-phase2/issues/75 (Upload coverage reports & make more visible), we're adding the reports (in addition to printing them in logs) to the Google Cloud Storage [hopic-test-coverage-reports bucket](https://console.cloud.google.com/storage/browser/hopic-test-coverage-reports;tab=objects?forceOnBucketsSortingFiltering=true&project=phx-01h4rr1468rj3v5k60b1vserd3&prefix=&forceOnObjectsSortingFiltering=false). The reports are saved under a branch-named folder, as json with filenames of <timestamp>-<commit-sha>.json.

(Note - chose json so we could easily pull out what we needed if reporting back somewhere like GitHub, but the [flag options are more limited than report](https://coverage.readthedocs.io/en/7.2.7/config.html#json) and it prints out excess lines.)

Unfortunately, as a result, cloudbuild.yaml now has additional steps as it seems you can use 'name' gcr.io/cloud-builders/gsutil, gcr.io/cloud-builders/docker or gcr.io/cloud-builders/gcloud, but not combine, so I separated the steps out (docker compose up, copy report from volume to gcs, tear down). If this is a potential issue we could look into combining the tests/ load/ tear down into one step by also downloading the tools needed (gcloud/ gsutil) in that same step, but that seemed messier.

The teardown step technically isn't required, but I think you mentioned on the last test PR that it's best practice. I was thinking of potentially loading the coverage report within the docker compose, but then that would required passing credentials and loading the gsutil component, and it then wouldn’t be reusable for running tests locally.

There was an issue with permissions to access the docker volume - tried a few iterations, but flagged it back to user:root in the docker-compose file. I imagine this will change. (I was trying out adding the dir in Dockerfile then giving permissions there, but I think I was doing it wrong, and not entirely sure what permissions it needs for cloudbuild. )

This also assumes that the google cloud storage API is enabled and that cloudbuild has read-write permissions.

TODO:
- [x] Get total coverage percentage from coverage.json and fail if below the threshold
- [x] Fail on main branch only
- [x] Retrieve last build (on all branches) and report on the differences in coverage
- [x] Create issue to write back test coverage and build status badge to GitHub 
-


cloudbuild.yaml
```diff
@@ -49,20 +50,46 @@ steps:
...
+      if (( $(echo "${total_coverage} < ${TEST_COVERAGE_THRESHOLD}" | bc -l) )); then
+        echo "Coverage (${total_coverage}%) is below the threshold of (${TEST_COVERAGE_THRESHOLD}%) - please add more unit tests!"
+        echo "For more details, read the test coverage report logged in 'Run tests' cloud build step."
+        # Fail build if main branch.
+        if [[ "${BRANCH_NAME}" == 'main' ]]; then  
+          echo "Failing build for insufficent test coverage."  
+          exit 1
+        fi
+      else
+        echo "Coverage (${total_coverage}%) is above the threshold. Continuing..."
```
Stephen-ONeil, line 71: I'd suggest also including the `TEST_COVERAGE_THRESHOLD` value in this output, just for reference.

cloudbuild.yaml
```diff
@@ -49,20 +50,46 @@ steps:
...
+      
+      # Compare total coverage to failure threshold.
+      total_coverage=$(jq '.totals.percent_covered' ./coverage/coverage.json | awk '{printf "%.2f\n", $1}')
+      if (( $(echo "${total_coverage} < ${TEST_COVERAGE_THRESHOLD}" | bc -l) )); then
+        echo "Coverage (${total_coverage}%) is below the threshold of (${TEST_COVERAGE_THRESHOLD}%) - please add more unit tests!"
+        echo "For more details, read the test coverage report logged in 'Run tests' cloud build step."
+        # Fail build if main branch.
+        if [[ "${BRANCH_NAME}" == 'main' ]]; then  
+          echo "Failing build for insufficent test coverage."  
+          exit 1
```
Stephen-ONeil, line 68: Instead of exiting immediately, I think it'd be better if exiting happened after all of the stats had been calculated and logged (coverage threshold _and_ deltas). That way the information to be found in the build output is consistent each time.

Since I think we might want additional failure states, which will not be mutually exclusive, I'd suggest tracking failure reasons in an array. Declare an empty bash array higher up with `failure_reasons=()` and append failure causes to it with `failure_reasons+=("Another reason this failed")`. You can check whether the array is empty with `if [ ${#failure_reasons[@]} -eq 0 ]; ...` at the end. If it's not empty, echo any items in it and then exit with a failure state.

cloudbuild.yaml
```diff
@@ -49,20 +50,46 @@ steps:
...
+        fi
+      else
+        echo "Coverage (${total_coverage}%) is above the threshold. Continuing..."
+      fi
 
-      docker compose -f docker-compose.run-tests.yaml down -v
+      # Find coverage delta between this and last commit.
```
Stephen-ONeil, line 74: I'd suggest using deltas between commits only while on the main branch (if we're squashing branches, this is effectively deltas from whole branches, which is decently useful). 

On non-main branches I'd suggest a delta against the most recent coverage report from main (branches off branches, where main isn't the correct baseline, shouldn't happen enough for us to bother accounting for them).

Stephen-ONeil (changes requested):
Nice! I've got some suggestions for now, but I think you, me, and @AlexCLeduc should also have a quick discussion about when and why to fail over test coverage. Alex's opinion matters most, since this is mostly about desired app code dev workflow.

Personally, I think main should possibly never fail due to low test coverage. Day to day, it's a bit inverted because test coverage _should_ be considered _before_ a branch is merged, not after. Assuming a GitOps flow where CI/CD on the main branch is how you deploy to prod (not that we've settled on that 100%), then operational concerns need to trump test coverage ideals and the code should deploy (assuming the actual tests and health checks still pass, that is).

I also think the test coverage delta is a more important reason to fail a branch (see my comment on the delta calculation too, I think the baseline coverage should be main rather than previous commit). It doesn't have to be strictly positive, but as long as we don't let a single branch cause too much back-sliding, and we keep testing coverage in mind while reviewing, it'll be a way to keep things on track while staying flexible for the reality of code that's not worth testing or operational deadlines that trump coverage targets. 

Having checks for both is really useful for early stage/projects under rapid development too. Not a concern on this project because the coverage is already pretty good, but if we use this pattern on a project that is starting off with a low total coverage then the strategy would be to set the absolute threshold target at the floor combined with a positive delta threshold, to nudge things towards a better state over time.

cloudbuild.yaml
LilaKelland: Implemented this - logging first, and not failing on main branch  (only if not-main branch (but also commented that out as you're right we may not want to have this fail.))

cloudbuild.yaml
LilaKelland: Deltas are calculated against main branch.

deploy/gcloud_env_vars.sh
```diff
@@ -23,6 +23,8 @@ export GITHUB_MAIN_BRANCH_NAME=main
 export PROJECT_IS_USING_WHITENOISE=true
 
 export TEST_COVERAGE_BUCKET_NAME=hopic-test-coverage-reports
+export TEST_COVERAGE_THRESHOLD=80
+export TEST_DELTA_THRESHOLD=-10  #current commit test coverage minus last commit on main's coverage
```
Stephen-ONeil, line 27: Now that I think about it, the delta threshold is something that kind of depends on the size of the code bast. Let's not bother getting fancy with it though, haha, it'll just be something to tweak over time.

I think the delta threshold could be lower though, there's not much room for a branch to decrease by 10% of the code base size and not _also_ end up below the coverage threshold of 80% of the code base. Maybe a value of -5 or even lower?

Stephen-ONeil (approved):
Just one minor comment from me! I'm approving otherwise (:tada:)

deploy/gcloud_env_vars.sh
LilaKelland: I wanted to give a large margin, but what you've mentioned makes sense and have changed it back to -5.

PR #102, opened 2023-08-01
Cloud SQL observability configuration (DB audit logging, improved OpenTelemetry tracing)

Stephen-ONeil:
Related to #84

TODO:
  - [x] configure Cloud SQL postgres database flags for better logging
    - related steps added to `gcloud_init_setup.sh`
    - one step remains a TODO, running `CREATE EXTENSION pgaudit;` inside the DB instance. See #84 
  - [x] add `Psycopg2Instrumentor` to the application for better DB tracing
  - [x] to deal with the growing performance impact of all this extra telemetry, switched from `SimpleSpanProcessor` to `BatchSpanProcessor`. `BatchSpanProcessor` uses a worker thread; using it safely with requires special consideration for both gunicorn and the Cloud Run environment
    - [x] for gunicorn, needed to move instrumentation in to a gunicorn `post_fork` hook, so that  `BatchSpanProcessor` and application instance worker threads would be 1-to-1. This prevents us from using `preload_app = True`, but it's a net performance win even without gunicorn preloads
    - [x] for Cloud Run, there are two big things to account for when running anything in a background process or worker thread:
      - [x] Cloud Run containers only get CPU time while handling a request. We specifically don't want to hold up responses to wait on telemetry processing, so we can't guarantee that the `BatchSpanProcessor` won't still have un-flushed traces on its worker thread when it's throttled off the CPU. The processes will continue work the next time a request comes in/CPU is allocated, so this may makes our traces eventual rather than instant. This isn't something we have to remediate ourselves, just something to be aware of
      - [x] what about `BatchSpanProcessor` that hasn't yet resolved when a container is fully killed, rather than just temporarily throttled?  GCP promises 10 seconds of grace-period CPU time for processes to respond to the `SIGTERM` signal and exit. To make use of this, I attach a `flush_telemetry_callback` to each spawned worker in our gunicorn `post_fork` hook, and call that function from our `worker_exit` hook. Gunicorn will receive and handle the `SIGTERM` and, unless our containers are massively slow/clogged up with exit tasks, there should be plenty of time during those 10 seconds to flush the `BatchSpanProcessor`(s) 
    - I figure eventual traces are better than dropped traces, and both are better than significantly slowing down request handling
    
For reference, here's an example of the benefits from the Trace Explorer perspective. 
image
The selected span in the trace below is from `Psycopg2Instrumentor`; we always get there for every DB query. The sub-spans of that (Cloud SQL Query, Aggregate, Index Scan) are generated from the DB's logs; this is enabled by the metadata included in each query statement by `sqlcommenter`, which Cloud SQL Query Insights will see in statement logs and use to construct after the fact trace information. Note that there's a limit there, Query Insights will only do this for up to the first 20 queries logged in a minute, and seems to often choose to sample from less.  

That's annoying, but it's not a big deal. You can grab the parent span ID from a `Psycopg2Instrumentor` span, or the overall trace ID for the request, and search for it in the log explorer to pull up all the fine details :+1:.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -14,6 +14,15 @@
       "justMyCode": true
     },
     {
+      "name": "gunicorn",
+      "type": "python",
+      "request": "launch",
+      "program": "${workspaceFolder}/venv/bin/gunicorn",
+      "cwd": "${workspaceFolder}/server",
+      "django": true,
+      "justMyCode": true
+    },
+    {
       "name": "test",
       "type": "python",
       "request": "launch",
--- a/deploy/gcloud_env_vars.sh
+++ b/deploy/gcloud_env_vars.sh
@@ -17,7 +17,7 @@ export PROJECT_REGION=northamerica-northeast1
 export BUILD_GITHUB_REPO_NAME=cpho-phase2
 export BUILD_GITHUB_REPO_OWNER=PHACDataHub
 
-export GITHUB_MAIN_BRANCH_NAME=main
+export GITHUB_MAIN_BRANCH_NAME=database-observability-configuration # main TODO revert this before merging
 
 # NOTE: if PROJECT_IS_USING_WHITENOISE is true then no cloud storage will be created for the project
 export PROJECT_IS_USING_WHITENOISE=true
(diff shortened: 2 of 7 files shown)


LilaKelland (approved):
Awesome - as I mentioned before - greatly appreciate the explanations here! Just looked up how to use .vscode/launch.json - neat.

PR #104, opened 2023-08-03
Add script to support the use of cloud-sql-proxy in dev

Stephen-ONeil:
Our DB has a private IP, accepting connections only inside the project's virtual private cloud. The Cloud Run app service connects to this with a Serverless VPC Connector. By default, we do not give the DB a public IP at all.

But, especially in the pre-prod phase, we still need to touch the Cloud SQL instance now and then. To enable this, I've added a script that wraps `cloud-sql-proxy` and adds a setup step that temporarily enables a public IP, allow listing only the external IP of the dev machine running the script, as well as an exit trap cleanup step that clears the allow list and disables the public IP. 

Using this script requires you to be logged in as (and re-oauth as, to generate a temporary credential file) a google account with sufficient IAM permissions for patching and connecting to the DB.

Note: this script just proxies the database to `localhost:6543`. The next step, of connecting a client to the DB and authenticating as a DB user, is up to the dev using it. That likely means having secret manager access to the DB credentials generate and saved during the `gcloud_init_setup.sh` script.

### DIFF
--- a/deploy/connect_cloud_sql_proxy.sh
+++ b/deploy/connect_cloud_sql_proxy.sh
@@ -13,15 +13,11 @@ if ! which cloud-sql-proxy; then
   exit 1
 fi
 
-echo ""
-echo "Enabling public IP for database instance, initially configured to refuse all public IP connections"
-gcloud sql instances patch "${DB_INSTANCE_NAME}" --clear-authorized-networks --assign-ip
-
-echo ""
-echo "Getting temporary credentials for cloud-sql-proxy (oauth step)"
-gcloud auth application-default login
-
 function cleanup {
+  echo ""
+  echo "Deleting temorary app env file"
+  rm -f "${local_access_env_path}"
+
   echo ""
   echo "Revoking temporary credentials used for cloud-sql-proxy"
   gcloud auth application-default revoke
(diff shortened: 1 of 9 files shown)


AlexCLeduc:
Is `localhost:6543` for real or is that a typo? Should we use a different env file?

Can we add some docs around how to connect and how to cleanup once we're done? Is this a process that runs until we kill it or do we need to run an explicit cleanup function?

Stephen-ONeil:
> Is `localhost:6543` for real or is that a typo?
Real, the default postgres `5432` is going to be in use on a dev's machine, it'd be annoying to kill your local postgres any time you need to connect to remote. 

> Should we use a different env file?
For connecting a local django app to the prod DB? Yeah, you'll need a `.dev.env-secret` file to overwrite the `DB_...` env vars, including a DB user it's _actually_ secret DB password. I'm considering adding another DB user specifically for this, so that we don't have to use the prod app DB user or root for this. Could do that before this PR merges, maybe save the whole local connection `.env.dev-secret` file as a GCP secret :shrug:.

> Can we add some docs around how to connect and how to cleanup once we're done? Is this a process that runs until we kill it or do we need to run an explicit cleanup function?
Any dev trusted to connect to the prod DB should know how to connect to the prod DB ;), haha, but ok. As for cleanup, this process runs until killed, and once killed it's exit trap performs the relevant cleanup. "Relevant" to this script doesn't include anything done to the local .env files, because this script is agnostic to _how_ you connect to the DB after starting the proxy. If that's part of the cleanup you'd want to see, I _could_ extend this script to assume django app connection and manage the necessary secrets, although I'm more wary about automating that.

AlexCLeduc:
Ahh ok, makes sense, a different DB user sounds like a good idea, if this ever leaks or something we don't have to reset as much 👍 


Aren't we already using `.env.dev-secret` for "local dev secrets" like `GC_NOTIFY_DEV_KEY` and `SOME_TRANSLATION_SERVICE_API_KEY` ? Wouldn't we want a 3rd file here? I'm open to having to use an explicit settings module for any commands, e.g. `manage.py --settings=settings.local_prod runscript ...`, and our `local_prod.py` settings file would essentially `from regular_settings import *` but overwrite DATABASES. If we don't want to complicate settings_utils, the local_prod could just read in DB connection details from a JSON file

What I like about this explicit settings CLI argument is that it's harder for someone to mess up and run something like `flushdb` in production

Stephen-ONeil:
I guess that third file could be a local `.env.prod` (with the dev-access prod DB user, and without the prod django secret, unnecessary API keys, etc). Do agree that having to explicitly add an alternate `--settings` would help make it harder to mess up prod though, yeah.

This was something I was thinking of doing for the GitOps based (data) migration flow, but maybe one of the set up steps in `connect_cloud_sql_proxy.sh` could be triggering a DB backup.

Stephen-ONeil:
Finally circled back to this. Added a second DB user and a second `.env.prod`-like file in Secret Manager, for use when connecting to the prod DB from a local dev environment. Fetching/cleaning up that file is part of the connection script.

Rather than the `--settings=settings.local_prod` approach suggested above, I just added a "y/n" input confirmation when the `settings.py` detects the local prod configuration. Could be either way though.

This could be a lot tighter in a lot of ways, but it's already more than sufficient for the current dev state of the app in my opinion. MEant to be temporary anyway so I won't bother polishing it further. Sooner rather than later, we'll probably be using `kubectl exec` to perform these operations from the cloud, rather than letting devs connect directly to the prod DB.

AlexCLeduc (approved):

Issue #65, opened 2023-07-14
Cloud Run uptime checks


Stephen-ONeil:
https://cloud.google.com/blog/products/management-tools/monitor-cloud-run-service-availability-with-uptime-checks

Cross posting from https://github.com/PHACDataHub/cloudrun-deployment-example/issues/24

Since we're using CPHO as our Cloud Run deployment experiment/example repo for now (because it'll force us to also consider the realistic dev experience and prod impacts of whatever we're doing with infrastructure), I might move more of these issues over and possibly archive that repo.

LilaKelland:
Click opt'd in to set this up.  Alerts are configured to @Stephen-ONeil's gcp email (mobile app) - lets see if this works!  Next step - see if we can add to set up scripts (I think there are ~ 4 API's to enable).

Stephen-ONeil:
I've updated the uptime alert to only happen every 15 minutes (900s), instead of every minute. Two reasons: 1) it was really clogging up my telemetry logs, and 2) that'd keep the Cloud Run container hot way more than it needs to be and _might_ have inflated the bill (dashboard billing estimate still says only about $1.50 for July, haha, so I don't think that's a worry so much).

Longer term, for point 1, we should have a dedicated uptime/health check URL defined in the app. Then I could target and disable trace accumulation for it.

For point 2, just something to keep an eye on when we have more accurate billing numbers.

Not super related, but here's a relevant blog post for health/uptime check best practices: https://philbooth.me/blog/six-ways-to-shoot-yourself-in-the-foot-with-healthchecks

Stephen-ONeil:
I've added an OpenTelemetry-ignored simple `/healthcheck` route (#105), going to update the uptime check to ping it.

Stephen-ONeil:
Closed by #106

PR #105, opened 2023-08-04
Simple healthcheck route

Stephen-ONeil:
A better route than `/` for the automatic GCP health checks to hit. Configured to be ignored by OpenTelemetry instrumentation too.

Related to #65.


Stephen-ONeil:
Trivial, self approving

PR #108, opened 2023-08-14
Implement PhacOrgs and update Rules

msarar:
PhacOrg & PhacOrgRoles tables added.
Tables populated on seed.
Indicator table updated along with indicator factory.

Rules updated to use phacOrgs.
Test created to test new rules.
User management page added.


server/cpho/models/phacorgs.py
```diff
@@ -0,0 +1,81 @@
...
+    if self.parent is None or self.parent_id == self.id:
+        return acronym
+    return f"{_get_phacorg_acronym(self.parent)}-{acronym}"
+
+
+@add_to_admin
+class PHACOrg(models.Model):
+    def __str__(self):
+        s = _get_phacorg_acronym(self, get_lang_code())
+        s += tdt("colon") + " "
```
AlexCLeduc, line 36: I think ": " would work just fine, no need to translate the colon, right now there's no space in between the acronym and the "colon"

server/cpho/jinja2/user_management/user_management_page.jinja2
```diff
@@ -0,0 +1,46 @@
...
+      {% for user in user_metadata %}
+        <tr>
+          <td>{{ user }}</td>
+          <td>
+            {% if user.last_login %}{{ date_display(user.last_login) }}{% endif %}
+          </td>
+          <td>
+            {% for role in user_metadata[user] %}
+              {# {% if role.is_lead %}<span class="fw-bold">{{ tdt("Lead: ") }}</span>{% endif %}
+                  {{ role.phac_org_name }} #}
```
AlexCLeduc, line 29: Some old code is left behind here

server/cpho/management/commands/seed_programs.py
```diff
@@ -0,0 +1,130 @@
...
+    skipped_count = 0
+    with open(BASE_DIR / "scripts/phac_orgs.csv", encoding="utf-8") as f:
+        reader = DictReader(f)
+        for data in reader:
+            data["pk"] = data["pk"] if data["pk"] != "" else None
+            data["parent_pk"] = (
+                data["parent_pk"] if data["parent_pk"] != "000000" else None
+            )
+
+            is_branch = int(data["pk"]) % 10000 == 0
```
AlexCLeduc, line 92: Couldn't this just be a `@property` or annotation? (An annotation has the advantage of being available for filtering)

My concern is we'll probably have to uproot this a bit when we start putting PHACOrg into its own helper package, better to not add any columns that aren't already in ITAP 

It's not a huge deal right now, though, it's easy enough to change later

server/tests/test_permissions.py
```diff
@@ -0,0 +1,167 @@
+import time
```
AlexCLeduc, line 1: It's good to test stuff without the mocks, but I think things will move quicker and smoother if you follow the pattern of testing authorization in two completely separate parts, 

1. Adding `patch_rules()` in ordinary request-based tests, making sure to test the 200 and 400 cases. This is not just a quick and dirty way to get tests passing (which you are doing), but also a way to ensure requests are rejected when the rule doesn't pass (which you aren't doing). It also insulates those tests from rule behaviour changes. 
2. `test_permissions.py` need only worry about testing rule implementations themselves, by running assertions against the non-mocked `test_rule` directly with concrete users, branches and indicators. It doesn't need to worry about views if we cover those in 1)

AlexCLeduc (approved):

PR #109, opened 2023-08-17
4982 kd export indicator

kdompaul:
Indicator data files can now be exported


server/cpho/views/view_util.py
```diff
@@ -130,6 +130,86 @@ def upload_mapper():
         },
     }
 
+def export_mapper():
```
AlexCLeduc, line 133: Is there anything we can re-use from Rad's CSV import tools? I would think there's duplication here

kdompaul: I was originally thinking for the export mapping to modify the function by reversing the keys and values in each dictionary within "Upload_mapper". So export mapper function takes the dictionaries generated by upload_mapper and creates new dictionaries with the keys and values reversed. But I wasn't sure if it was a good idea to make it dependent on upload mapper which is why i did it the way i did but let me know if you think i should follow through with this

def export_mapper():
upload_mapper = upload_mapper()
export_mapping = {
category_mapper: {v: k for k, v in upload_mapper["category_mapper"].items()},
etc..
}
here's like a brief idea of how I would implement it

AlexCLeduc: I think it's a good idea to use the same key/values here. If one changes, it's more likely we want the other to change than not. If we have an exceptional key/value pair, we can always override it anyway

msarar (changes requested):
image


Upon downloading and then uploading a file I got this error
rename "Value_Unit" to "Value_Displayed" in the csv header

Also, some Dimension Values are empty
image

msarar (approved):

AlexCLeduc (approved):

PR #111, opened 2023-08-22
5021 rs improve import

msarar:
Data in imported files only get saved in the DB if the data in the file is different than the data in the DB. 
This results in not overwriting approved data if there's no change

Also, Upload did not have permission enforcements. This branch fixes that.


server/cpho/views/upload_indicators.py
```diff
@@ -203,6 +254,38 @@ def clean_csv_file(self):
...
+            # checking if indicator already exists
+            indicator_obj = Indicator.objects.filter(
+                name=data_row["Indicator"],
+                category=mapper["category_mapper"][data_row["Category"]],
+                sub_category=mapper["subcategory_mapper"][data_row["Topic"]],
+                detailed_indicator=data_row["Detailed Indicator"],
+                sub_indicator_measurement=data_row[
+                    "Sub_Indicator_Measurement"
+                ],
+            )
```
AlexCLeduc, line 267: Rather than testing the length twice and calling `[0]` (3 queries), just call `indicator_obj = Indicator.objects.filter(...).first()` and test for `None` rather than the length (1 query)

server/cpho/views/upload_indicators.py
```diff
@@ -30,48 +36,80 @@ class UploadForm(forms.Form):
...
+                if datum["Value_UpperCI"] != ""
+                else None
+            ),
+            value_unit=mapper["value_unit_mapper"][datum["Value_Displayed"]],
+            single_year_timeframe=datum["SingleYear_TimeFrame"],
+            multi_year_timeframe=datum["MultiYear_TimeFrame"],
+        )
 
+        if len(indData_obj) == 0:
+            print("Editing Indicator Data")
```
AlexCLeduc, line 112: Look through and find undeleted print statements

AlexCLeduc (approved):
Thanks Rad!

PR #114, opened 2023-08-25
Use docker build arg to select between prod and dev-test dependency sets

Stephen-ONeil:
Nearly trivial, but it might be a pattern worth reusing so I'm making a PR for visibility.


AlexCLeduc (approved):

LilaKelland (commented):
Awesome!

PR #116, opened 2023-08-29
add period to import and export

msarar:
Period info can be added during import/export.

Works with prover versioning too :)

### DIFF
--- a/server/cpho/views/upload_indicators.py
+++ b/server/cpho/views/upload_indicators.py
@@ -62,13 +62,14 @@ class UploadForm(forms.Form):
 
     def handle_indicator_data(self, indicator_obj, datum):
         mapper = upload_mapper()
-        default_period = (
-            Period.relevant_years().filter(year_type="calendar").first()
-        )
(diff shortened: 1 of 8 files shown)


server/cpho/views/upload_indicators.py
```diff
@@ -63,9 +62,13 @@ def handle_indicators(self, datum):
...
         mapper = upload_mapper()
-        preiod_val = Period.objects.get(
-            year=2021, quarter=None, year_type="calendar"
+        default_period = (
+            Period.relevant_years().filter(year_type="calendar").first()
         )
+        period_val = default_period
+
+        if "Period" in datum and datum["Period"] != "":
+            period_val = mapper["period_mapper"][datum["Period"]]
```
AlexCLeduc, line 71: If we're gonna support multiple-periods we should probably force everyone to use the column, especially since it saves records automatically without showing them what it does beforehand. So let's not offer a default at all

AlexCLeduc (approved):

PR #117, opened 2023-08-29
add modify/create user views, use real groups for hso/admin/account-managers

AlexCLeduc:
Mostly copying ITAP but did it quick and dirty because we'll be canning parts of it later when we (probably) move to indicator-based authorization

### DIFF
--- a/server/cpho/util.py
+++ b/server/cpho/util.py
@@ -6,11 +6,7 @@ from django.utils.translation import get_language
 
 import pytz
 
-from cpho.constants import (
-    ACCOUNT_MANAGER_GROUP_NAME,
-    ADMIN_GROUP_NAME,
-    HSO_GROUP_NAME,
-)
+from cpho.constants import ADMIN_GROUP_NAME, HSO_GROUP_NAME
 
 eastern_timezone = pytz.timezone("Canada/Eastern")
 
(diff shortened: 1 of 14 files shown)


server/cpho/views/user_management.py
```diff
@@ -47,3 +65,197 @@ def get_context_data(self, **kwargs):
...
+    def get_form(self):
+        if self.request.POST:
+            return CreateUserForm(self.request.POST)
+        else:
+            return CreateUserForm()
+
+    def get_or_create_user(self, form):
+        email = form.cleaned_data["email"]
+        return User.objects.create_user(
+            email=email, username=email, password=email
```
AlexCLeduc, line 223: we can stick with password=username=email for user testing

server/cpho/util.py
```diff
@@ -22,3 +30,25 @@ def group_by(iterable, key):
...
+
+    @classproperty
+    @lru_cache
+    def account_manager_group(cls):
+        return Group.objects.get_or_create(name=ACCOUNT_MANAGER_GROUP_NAME)[0]
+
+    @classproperty
+    @lru_cache
+    def hso_group(cls):
+        return Group.objects.get_or_create(name=HSO_GROUP_NAME)[0]
```
AlexCLeduc, line 54: trying something new here, it's a neat shorthand and involves at most 3 queries, it'll even get cached across multiple requests (which is OK because these are totally static)

server/cpho/views/user_management.py
msarar: I agree that for now we can keep username=passowrd
but just wanted to discuss if username= email is the best idea..
as of iright now we have other users like cdsb_lead, cdsb_user etc from the seed script, whose username does not end with @canada.ca etc.
Could this create confusion?

msarar (approved):
Thanks Alex :)

PR #119, opened 2023-08-31
add user-changelog, redirect to it from successful uploads

AlexCLeduc:
I thought it would be useful for users to see their changes when they upload a CSV. Unfortunately that is complex but what we can easily do is show them a changelog of all their changes EVER, and naturally the upload changes will be the first changes they see 
Screenshot 2023-08-31 at 16 30 50

### DIFF
--- a/server/cpho/jinja2/base.jinja2
+++ b/server/cpho/jinja2/base.jinja2
@@ -97,6 +97,11 @@
                       <i class="bi bi-power"></i>
                       {{ tm("logout") }}
                     </a>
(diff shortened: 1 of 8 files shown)


msarar (approved):
image
LGTM!
This is super useful

One thing i wanted to discuss:
would it be useful to break the "this change"
column to two: "this change value" and "this change type"
Only bringing it up as newly created indicators/indicatordatum objects only show created.

Maybe an idea for the future.

PR #125, opened 2023-09-26
Cloudbuild k8s support

Stephen-ONeil:
Minor tweaks to allow cloud build to still build and push images on the main branch, even while `GITHUB_MAIN_BRANCH_NAME` (and therefore cloud run deploys) are locked to the user-testing branch.  Update image tags and flux configuration, so that flux will update k8s on image push.


vedantthapa (approved):
Since this branch was a fork of [simsingh/k8s-infra](https://github.com/PHACDataHub/cpho-phase2/tree/%40simsingh/k8s-infra), I've added the remaining commits from the branch to do some additional k8s configurations, add an architecture diagram and re-structure the `architecture-diagram/icons/` directory a bit (made changes to the existing d2 diagrams, so nothing should be broken).

PR #126, opened 2023-09-28
feat: add graphQL API

AlexCLeduc:
Added the general framework and all the tools (testing utils, dataloaders) I think we'll need to create the API infobase will need. I added a complex dataloader I think we can re-use and tests. 

This PR introduces the concept of dataloaders. These are batching tools that work in mysterious ways, based on promises, which simulate async behaviour. Without dataloaders, we would have really bad unsolvable N+1 queries. Note that there are layers of complicated concepts here,

1. Promises - I don't understand how these work under the hood and you probably don't need to either, it's just a magic thing that allows lazy evaluation
2. GraphQL promise execution context plugin - Also don't understand how this works, you don't need to either
3. [Pleasant promise generator syntax](https://alexcleduc.com/posts/python-pleasant-promises/) - this is something I made, you not important if you just pretend it's async
4. dataloaders - This is a GraphQL concept, resolvers are executed independently. Dataloaders, via async or promises, allow resolvers to return lazy values that get fetched in batches. The promise/dataloader library is based on promises.
5. Dependency-injected Dataloader (`SingletonDataloader`) - My custom wrapper around dataloader. Plain dataloaders aren't composable and are annoying to set up in a central place. These classes flip the paradigm so instead of creating a central dict of dataloader instances, you pass the dict to dataloader constructors and it will return existing instances if they exist on the dict and register new ones if they don't. [The concept is explained here](https://alexcleduc.com/posts/graphql-dataloader-composition/) (not super important to understand)

I don't think we'll have to write that many dataloaders or use `yield` very often. 

To play around with the API, just visit `localhost:8000/graphiql`, here's a complex query showcasing what graphql can do, it might not return much if you don't have any properly submitted data. Your django toolbar should also report an accurate number of queries, thanks to the graphiql_debug_toolbar package I've added. This is important because graphql makes it very easy to end up with really bad n+1 problems
```gql
query {
  indicators{
    name
    id
    detailedIndicator
    subIndicatorMeasurement    
    data(year:2022){
      value
      literalDimensionVal
      dataQuality
      reasonForNull
      valueUnit
      valueDisplayed
      dimensionValue {
        id
        value
        nameEn
      }
      dimensionType {
        id
        nameEn
      }
​
      period {
        year
        quarter
        yearType
      }
    }
  }
}
``` 

I don't have any concrete plans to use the graphql schema internally to the data-collection app, but it might be a good idea at some point, especially if they want fancy dashboards with lots of relationship traversal.


server/server/urls.py
```diff
@@ -24,6 +26,16 @@
...
+    path(
+        "graphiql/",
+        csrf_exempt(api_views.GraphiQLView.as_view()),
+        name="graphiql",
+    ),
+    path(
+        "graphql/",
+        csrf_exempt(api_views.GraphQLView.as_view()),
+        name="graphql",
+    ),
```
AlexCLeduc, line 38: I moved these from cpho/urls to server/urls, it's cleaner if cpho/urls only covers the data-collection app

server/api/views.py
```diff
@@ -0,0 +1,72 @@
...
+from graphql_core_promise import PromiseExecutionContext
+from pleasant_promises.graphene import promised_generator_middleware
+
+from server.middleware import AllowUnauthenticatedMixin
+
+from .schema import schema
+from .util import GraphQLContext
+
+
+class CustomGraphqlView(BaseGraphQLView):
```
AlexCLeduc, line 14: I've had to override a lot of this view to get promises working properly.

server/api/views.py
```diff
@@ -0,0 +1,72 @@
...
+                f"{error_to_log.__class__.__name__}: {error_to_log}"
+            )
+            self.logger.error(traceback_str)
+
+    def get_context(self, request):
+        dataloaders = {}
+        return GraphQLContext(dataloaders, user=self.request.user)
+
+
+class GraphQLView(CustomGraphqlView, AllowUnauthenticatedMixin):
```
AlexCLeduc, line 56: Note the mixin, this can be accessed without authentication by infobase clients

PR #129, opened 2023-10-03
k8s pod and app process 1 to 1

Stephen-ONeil:
Experimental k8s tuning, to see how the application performs when limited to one django server process per pod. Each pod is using the minimum pod-total resources for GKE Autopilot. 

This is probably not ideal, but I think it's worth trying just to learn more to inform future deployment tuning. We'll have to do some load testing against this once deployed, but there's plenty of tools for that.

### DIFF
--- a/k8s/application/server/deployment.yaml
+++ b/k8s/application/server/deployment.yaml
@@ -22,8 +22,9 @@ spec:
         # these are overrides for the default configuration of those injected proxy sidecars
         resources:
           requests:
-            cpu: 100m
-            memory: 128Mi
+            # Experiment: using the minimum pod resources https://cloud.google.com/kubernetes-engine/docs/concepts/autopilot-resource-requests#compute-class-min-max
+            cpu: 100m # + `server` container = 250m
+            memory: 128Mi # + `server` container = 512MiB
       - name: server
         securityContext:
           runAsUser: 5678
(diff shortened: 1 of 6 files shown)


k8s/application/server/deployment.yaml
```diff
@@ -52,6 +61,7 @@ spec:
         resources:
           # Autopilot only considers requests (see https://cloud.google.com/kubernetes-engine/docs/concepts/autopilot-resource-requests#resource-limits)
           requests:
-            cpu: 250m
-            memory: 512Mi
+            # Experiment: using the minimum pod resources https://cloud.google.com/kubernetes-engine/docs/concepts/autopilot-resource-requests#compute-class-min-max
+            cpu: 150m # + `isito-proxy` container = 250m
```
vedantthapa, line 65: I did try setting these values to something less than the specified ones but autopilot would just bump it up during `kubectl create`. In this case, however, explicitly setting `istio-proxy` resource configs might help.

k8s/application/server/deployment.yaml
```diff
@@ -16,6 +16,15 @@ spec:
         app: server
     spec:
       containers:
+      - name: istio-proxy
+        image: auto
+        # istio-proxies are injected by anthos, see `istio.io/rev: asm-managed` in ../namespace.yaml
+        # these are overrides for the default configuration of those injected proxy sidecars
+        resources:
```
vedantthapa, line 23: If this doesn't work, another way to set this is using annotations: https://istio.io/latest/docs/reference/config/annotations/

vedantthapa (approved):

PR #131, opened 2023-10-04
feat: Relevant Dimensions

msarar:
In order to reduce the number of unnecessary dimensions, added a m2m field called "relevant dimensions" in indicator models.

During View/Edit/Review of indicator data for a certain period, the dimensions that show up depending on:
- if dimensions is checked as a relevant dimension for the indicator
- Dimensions that not checked as relevant only show up if they already contain data


server/cpho/queries.py
```diff
@@ -81,3 +81,11 @@ def get_submission_statuses(indicator, period):
         "statuses_by_dimension_type_id": submission_status_by_dimension_type_id,
         "global_status": global_status,
     }
+
+
+def relevant_dimension_types_for_period(indicator, period):
+    data = indicator.data.filter(period=period).distinct("dimension_type")
+    dimension_types = set([d.dimension_type for d in data]).union(
+        indicator.relevant_dimensions.all()
+    )
```
AlexCLeduc, line 90: I think we should have a `prefetch_related('dimension_type')` on the first query? If it doesn't play well with the `distinct()`, it's not a big deal because we're putting it through a `set()` anyway

server/cpho/views/indicator_data.py
```diff
@@ -396,7 +415,10 @@ def submission_status(self):
     def get_context_data(self, **kwargs):
         if self.dimension_type is None:
             predefined_dimension_types = DimensionType.objects.filter(
-                is_literal=False
+                is_literal=False,
+                code__in=relevant_dimension_types_for_period(
+                    self.indicator, self.period
+                ),
```
AlexCLeduc, line 421: Haven't tested this, but does this work considering we're comparing `DimensionType` records against `code` strings?

AlexCLeduc (approved):
LGTM, consider suggested changes

PR #132, opened 2023-10-06
K8s istio ingress cloud armor WAF configuration

Stephen-ONeil:
Add a backend, with the old WAF policy, to the istio ingress. I might have something wrong in here, someone should adopt this PR from me to double check and test it out.

### DIFF
--- a/k8s/istio/backendconfig.yaml
+++ b/k8s/istio/backendconfig.yaml
@@ -1,0 +1,11 @@
+apiVersion: cloud.google.com/v1
+kind: BackendConfig
+metadata:
+  name: istio-ingressgateway-backendconfig
+  namespace: istio-ingress
+spec:
+  enableLogging: "true"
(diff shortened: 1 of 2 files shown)


vedantthapa:
> While applying backendconfig.yaml in the k8s env, it throws decoding error for "spec.enableLogging". Can you check and confirm that the filed exists. Or can you share the link for documentation as well?

I think it might be:
```
spec:
  logging:
    enable: true
```
instead of https://github.com/PHACDataHub/cpho-phase2/blob/5dd83ebf2f653c3d620c74e50db1295b1509cff6/k8s/istio/backendconfig.yaml#L7

Reference: https://cloud.google.com/kubernetes-engine/docs/how-to/ingress-configuration#http_logging

simardeep1792 (approved):
Changes Checked and Approved!

PR #140, opened 2023-10-10
K8s prod env assumptions

Stephen-ONeil:
Our Cloud Run deployments got prod env vars from GCP secret manager and saved them to a `.env.prod` file. `config_util.get_project_config` use the existence of this `.env.prod` file to tell whether it should or should not try to read from the committed `.env.dev-public` file.

In k8s, we are injecting prod env vars directly in to containers as OS env vars. There's no longer a `.env.prod` file, so `config_util.get_project_config` was falling back to reading from `.env.dev-public`. With `decouple.Config`, this OS env vars are given priority, so the injected env vars were still used BUT any dev-only config values from `.env.dev-public` were also being surfaced. This mix of env vars disabled/broke certain prod-only behaviour.

Now, instead of looking for a `.env.prod` file, `get_project_config` checks for an OS env var `IS_K8S` that I'm setting in k8s, and doesn't use any `.env` files if it's set, assuming instead that all the env vars are already set in the OS env.

This should fix the k8s app server logs not being JSON formatted. Should also fix the lack of OpenTelemetry coming out of k8s.


Stephen-ONeil:
Any better way to detect when the app is executing inside of k8s?

vedantthapa (approved):
I recently pushed some changes to main for re-organizing resources in `k8s/`. I've merged those changes here too.

Everything else looks good to me!

PR #141, opened 2023-10-11
Refactoring the CPU and Memory usage for Hopic Project

simardeep1792:
Issue: GCP Kubernetes Cluster requests 1CPU and 1GB Memory which lead to high deployment cost.
Fix: Each deployment, pod and cluster has been modified and limits are being added to the allocation to memory and cpu.

### DIFF
--- a/k8s/application/postgres-sql/cnpg-cm.yaml
+++ b/k8s/application/postgres-sql/cnpg-cm.yaml
@@ -5,4 +5,4 @@ metadata:
   namespace: cnpg-system
 data:
   INHERITED_LABELS: app
-  INHERITED_ANNOTATIONS: proxy.istio.io/*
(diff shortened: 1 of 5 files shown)


vedantthapa:
Also, just to note it down; annotations are used instead of kustomize patches for postgres cluster at `k8s/application/postgres-sql/postgres-cluster.yaml` because the pods are provisioned via a cluster manifest.

vedantthapa (approved):
Also, just to note it down for future ref; annotations are used instead of kustomize patches for postgres cluster at `k8s/application/postgres-sql/postgres-cluster.yaml` because the pods are provisioned via a cluster manifest.

simardeep1792:
Squashed and Merged!

PR #142, opened 2023-10-12
Re-arrange k8s resources

vedantthapa:
- The `k8s/` directory is divided into sub-directories as per k8s namespaces. For instance, all resources under the`cnpg-system` namespace are located in the `k8s/cnpg-system/` sub-directory.
- The sync configurations for flux are decoupled from the flux-system directory. That is, every namespace now has it's own flux sync configurations stored in the `sync.yaml` file.
- Dependencies between resources are specified using the `dependsOn` key in each `sync.yaml` to ensure reproducible deployments and handle scenarios where, for instance, crds are required to be ready before a cr is deployed. The current order is:
    - flux-system
    - istio-ingress
    - cert-manager/crds
    - cert-manager/resources
    - cnpg-system
    - server/postgres
    - server/django
    
> Flux is suspended until this PR is merged.


simardeep1792:
Looks good to me , Thanks for the refactoring

simardeep1792 (approved):
LGTM!

PR #143, opened 2023-10-16
Configure `readinessProbe` for server

### LINKED ISSUE #139: Configure Istio HTTP health checks
As per these [docs](https://istio.io/latest/docs/ops/configuration/mesh/app-health-check/#liveness-and-readiness-probes-using-the-http-request-approach), we can configure HTTP based health checks via Istio for readiness/liveliness of our nodes. The app provides a simple health check route at `/healthcheck`, use that. Tune the initial delay and period; delay will probably be relatively stable going forward but the appropriate period may depend a lot on whether we stick with low-resource, single django process, nodes (related: #133).

vedantthapa:
Add readiness probe at `/healthcheck` endpoint.

Closes #139

### DIFF
--- a/k8s/server/django/deployment.yaml
+++ b/k8s/server/django/deployment.yaml
@@ -14,6 +14,8 @@ spec:
     metadata:
       labels:
         app: server
+      annotations:
+        sidecar.istio.io/rewriteAppHTTPProbers: "false"
     spec:
       containers:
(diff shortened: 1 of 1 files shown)


Stephen-ONeil (commented):
LGTM

Stephen-ONeil (approved):

PR #144, opened 2023-10-16
Configure alerts for flux

vedantthapa:
Configure flux to send deployment alerts to the slack channel.

Currently the alert severity is set to `info`, however, it can be set to `error` to reduce noise. [Here's](https://fluxcd.io/flux/monitoring/alerts/#define-an-alert) more info on these options. Alternatively, we can also fine-tune the configuration to [include](https://fluxcd.io/flux/components/notification/alerts/#event-inclusion) / [exclude](https://fluxcd.io/flux/components/notification/alerts/#event-exclusion) certain events for individual `eventSources` via regex patterns.


simardeep1792:
I think the alert severity can be reduced to Error considering we are receiving the alerts on Slack quite frequently

simardeep1792 (approved):
LGTM!

vedantthapa:
In addition to the slack channel updates, I've added the configuration to update the github commit status. Currently the `eventSeverity` for this configuration is set to `error` for all kustomization resources.

It probably won't work unless an org admin approves my request for a personal access token for this repo. I, therefore, suggest to keep this PR open until we get an approval and can verify that things are working as expected.

Stephen-ONeil (approved):
Nice, thanks!

PR #147, opened 2023-10-23
feat: Upload preview

msarar:
Upon upload, file data gets stored into session storage
the preview upload function displays the data from the session
and then saves the data to the backend.

### DIFF
--- a/server/cpho/jinja2/indicators/upload/preview_upload.jinja2
+++ b/server/cpho/jinja2/indicators/upload/preview_upload.jinja2
@@ -1,29 +1,54 @@
 {% extends 'base.jinja2' %}
+{% from 'cpho_macros.jinja2' import check_error, check_error_class %}
 {% block content %}
 
-  <a href="#" hx-post="{{ url("save_upload") }}" class="btn btn-primary">{{ tdt("Confirm") }}</a>
-  <a href='' class="btn btn-danger">{{ tdt("Cancel") }}</a>
 
-
+ 
+  <a id='upload_btn'
+     href="#"
+     onclick="showLoading()"
+     hx-post="{{ url("save_upload") }}"
+     class="btn btn-primary {{ disabled if not no_errors }}">{{ tdt("Upload Data") }}</a>
+  <button id='loading_btn'
+          class="btn btn-primary d-none"
+          type="button"
+          disabled>
+    <span class="spinner-grow spinner-grow-sm" role="status" aria-hidden="true"></span>
+    {{ tdt("Uploading Data") }}
+  </button>
+  <a href="{{ url("list_indicators") }}" class="btn btn-danger">{{ tdt("Cancel") }}</a>
+ 
   {% for indicator in grouped_data %}
     <div class='mt-5 card'>
       <div class="card-header">
-        <div class="h5">{{ tdt("Indicator") }} : {{ indicator[0] }}</div>
+        <div class="h5">
+          {{ tdt("Indicator") }} : {{ indicator[0] }}
+          {% if grouped_data[indicator][0]["new_indicator"] %}<span class="new-ind-tag">{{ tdt("New") }}</span>{% endif %}
+        </div>
  
-      </div>
 
+      </div>
       <div class=" card-body">
+
+        {{ (check_error(grouped_data[indicator][0], "Indicator") ) }}
+ 
+
         <div class="p">
           <span class="fw-bold">{{ tdt("Detailed Indicator") }}</span> : {{ indicator[1] }}
         </div>
+ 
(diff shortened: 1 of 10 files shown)


server/cpho/jinja2/indicators/upload/preview_upload.jinja2
```diff
@@ -0,0 +1,132 @@
...
+     onclick="showLoading()"
+     hx-post="{{ url("save_upload") }}"
+     class="btn btn-primary {{ disabled if not no_errors }}">{{ tdt("Upload Data") }}</a>
+  <button id='loading_btn'
+          class="btn btn-primary d-none"
+          type="button"
+          disabled>
+    <span class="spinner-grow spinner-grow-sm" role="status" aria-hidden="true"></span>
+    {{ tdt("Uploading Data") }}
+  </button>
```
Stephen-ONeil, line 18: Hm, few issues here.

First, there's multiple accessibility issues with "disabling" a link through this `a.disabled` class. First, its styling information doesn't do anything to communicate info to a screen reader user. Second, `pointer-events: none;` doesn't stop keyboard events, so anyone using keyboard navigation can still "click" this link. You _could_ use the `aria-disabled="true"` attribute instead, to tell screen readers it's disabled, and change the CSS selector to `a[aria-disabled="true"]`, but that doesn't fix the keyboard interaction issue.

The proper fix is to not use an `a` here at all. Only use an anchor tag if the purpose is actually navigation. Accessibility aside, there's [lots of other issues](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#onclick_events) with using an `a` with `href="#"` in place of a `button`.

Additionally, if you use a `button` instead, then all you need to do to both prevent _any_ interaction, and to provide a11y support, is to add the [`disabled` attribute ](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attributes). Bootstrap even already has styles for `button[disabled="true"]`.

The second issue is this `#loading_btn` button. This one's more minor, but a `button`, even with the `disabled` and `role="status"` attributes, shouldn't be used for communicating status. It's weird [html semantics](https://developer.mozilla.org/en-US/docs/Glossary/Semantics#semantics_in_html), so you're fighting against the browser to make it work (having to `disable` the button behaviour). Better to just use a `span` with the bootstrap button styling (the `role="status"` is good though, leave that for sure :+1: )

server/cpho/jinja2/indicators/upload/upload_indicator.jinja2
```diff
@@ -0,0 +1,52 @@
...
+          <input type="submit"
+                 id="submit-btn"
+                 class="btn btn-primary mt-3 float-end"
+                 value="{{ tdt("Preview file data") }}">
+        </div>
+        <div id="loadingModal" style="display: none;">
+          <button class="btn btn-primary mt-3 float-end" type="button" disabled>
+            <span class="spinner-grow spinner-grow-sm" role="status" aria-hidden="true"></span>
+            {{ tdt("Getting data ready for preview") }}
+          </button>
```
Stephen-ONeil, line 33: See comment on `button#loading_btn` in `preview_upload.jinja2`.

Stephen-ONeil (changes requested):
Heads up, I tend to do thorough, almost pedantic, reviews :sweat_smile:. Especially when I'm reviewing for a new dev. I want to be clear, my goal with code review is as much about sharing knowledge as it is about getting the code ready to merge!

If you have time, read through any given comment to the end and follow any links (to skim or use as reference, at least) before making related changes in your code. I sometimes present multiple alternative solutions so I can cover more topics, instead of just jumping right to suggesting a "fix" :+1:.

Reach out any time if you have questions!

Bugs:
- [ ] preview table isn't responsive, overflows on narrow screens
image
  - probably just calls for an `overflow-x: scroll` on the parent of the table element. Note there's [further steps](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow#accessibility_concerns) to make sure the scrollable area is accessible after that
- [x] while testing this out locally, I found that uploading a sheet with an invalid value in the "Reason_for_Null_data" column threw an error. Look in to that and double check that there's no other columns with similar error checking bugs

Other:
- [ ] the preview page itself and needs tests. If you want suggestions on writing these let me know. Try to at least test that:
  - errors in all columns are handled and displayed
  - if errors exist then there is no active upload button
  - if no errors exist then there's an active upload button, and uploading is functional

msarar:
Thank you @Stephen-ONeil for taking the time to go so thoroughly through the code.
I appreciate your input :)

Stephen-ONeil (approved):
Looks good, thanks!

PR #149, opened 2023-10-30
Submission progress bar

msarar:
Added progress bar to indicate submission status

### DIFF
--- a/server/cpho/jinja2/indicators/_progress_bar.jinja2
+++ b/server/cpho/jinja2/indicators/_progress_bar.jinja2
@@ -29,6 +29,7 @@
       {{ tdt("Data entered") }}
       {% if not progress_in_table %}
         <i tabindex="0"
+           role="tooltip"
            class="bi bi-info-circle mx-1"
            data-bs-placement="bottom"
            data-bs-toggle="tooltip"
@@ -51,6 +52,7 @@
       {{ tdt("Program Approved") }}
       {% if not progress_in_table %}
         <i tabindex="0"
+           role="tooltip"
            class="bi bi-info-circle mx-1"
            data-bs-placement="bottom"
            data-bs-toggle="tooltip"
@@ -73,6 +75,7 @@
       {{ tdt("HSO approved") }}
       {% if not progress_in_table %}
         <i tabindex="0"
+           role="tooltip"
            class="bi bi-info-circle mx-1"
            data-bs-placement="bottom"
            data-bs-toggle="tooltip"
--- a/server/cpho/jinja2/indicators/_progress_bar.jinja2
+++ b/server/cpho/jinja2/indicators/_progress_bar.jinja2
@@ -14,8 +14,7 @@
 {% endif %}
 
 <div class='row mt-1 gx-2 {% if not progress_in_table %}mb-4{% endif %}'>
-  <div id="{{ indicator.id }}-{{ period.code }}-data-entered"
-       class="col-md-4 col-sm-12">
+  <div class="col-md-4 col-sm-12">
     <div class="progress" style="height:0.75rem">
       <div class="progress-bar bg-warning progress-data-entered"
            role="progressbar"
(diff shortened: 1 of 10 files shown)


Stephen-ONeil (changes requested):
Awesome to see you tackling the accessibility side of this one so thoroughly!

Have you tried it out with a screen reader your self? If not, and if you have access to one in your dev environment (PHAC windows laptop?), try navigating the status bar with it and let me know how it works for you. There were some things that could be improved when I checked with one my self, so if you've not used a screen reader before then this could be a good chance to get familiar with one :+1:.

I'll leave my own PR comments and suggestions once you've taken a stab at fixing any screen reader issues/pain points you can find your self.

If you can't get access to one let me know, we can get on a call some time and I can show you how the progress bar is read out by it for me.

server/cpho/jinja2/indicators/_progress_bar.jinja2
```diff
@@ -0,0 +1,86 @@
...
+    </div>
+    <div class="d-flex justify-content-center {% if not progress_in_table %}mt-1 fw-bold{% endif %}">
+      {{ tdt("Data entered") }}
+      {% if not progress_in_table %}
+        <i tabindex="0"
+           role="tooltip"
+           class="bi bi-info-circle mx-1"
+           data-bs-placement="bottom"
+           data-bs-toggle="tooltip"
+           data-bs-title="{{ tdt("Data currently in database but not approved by program or HSO. Data is not made visible through the API (for infobase) until approved by program and then HSO. Data modified after approval is also treated similarly") }}"></i>
```
Stephen-ONeil, line 36: An [`<i>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/i) doesn't seem to make sense here. It indicates "text that is set off from the normal text", which generally just means it's displayed in italics (and a slim chance that some screen readers my read it with an alternate voice). The only text inside this tag is the info symbol character though, added by the `bi-info-circle` class. The symbol itself doesn't have an italicized version, and a screen reader just reads it out as its code "f431" anyway.

The `<i>` itself isn't really a problem, just a weird use of it. The screen reader pronouncing "f431" is more of a problem.

You'll see in my suggested fix that I replaced the `<i>` with two divs. An outer one with the bootstrap tooltip attributes (and an `aria-label`) and an inner one for the icon. I did this so I could mark the icon, and it's "f431" content, as `aria-hidden` since it is just a visual indicator to tell visual users to mouse over it. The equivalent for a screen reader user is an `aria-label` indicating that the element is for accessing "more info", in addition to the `tabindex="0"` which prompts the screen reader to direct the user that there's an interactive element they can tab to within the region it's currently reading.

Stephen-ONeil (changes requested):
This is one of those cases where instead of giving suggestions, I had to test out solutions my self to come up with a fix. This should address the label and tooltip screen reader issues, see the individual comments on the equivalent code for explanations.

```
  <div class="col-4">
    <div class="progress" style="height:0.75rem">
      <div class="progress-bar bg-warning progress-data-entered"
        role="progressbar"
        style="width: {{ data_entered_percent }}%"
        aria-valuenow="{{ data_entered_percent }}"
        aria-valuemin="0"
        aria-valuemax="100"
        aria-labelledby="{{ indicator.id }}-{{ period.code }}-data-entered"
      ></div>
    </div>
    <div class="d-flex justify-content-center {% if not progress_in_table %}mt-1 fw-bold{% endif %}">
      <div id="{{ indicator.id }}-{{ period.code }}-data-entered">
        {{ tdt("Data entered") }}
      </div>
      {% if not progress_in_table %}
        <div
          tabindex="0"
          aria-label="{{ tdt("More info") }}"
          data-bs-placement="bottom"
          data-bs-toggle="tooltip"
          data-bs-title="{{ tdt("Data currently in database but not approved by program or HSO. Data is not made visible through the API (for infobase) until approved by program and then HSO. Data modified after approval is also treated similarly") }}"
        >
          <div class="bi bi-info-circle mx-1" aria-hidden="true"></div>
        </div>
      {% endif %}
    </div>
  </div>
```

Give it a try with navigator and let me know how it works. It's reasonable with my screen reader but mileage can vary between screen readers when not sticking to things that are very clearly defined in the specifications (in the specifications, tooltips seem to be a bit of an after thought and, in particular, it seems to be assumed that they'll be attached to interactive elements as a different way to do labels, rather than stand alone definition content).

If it's still weird then there's other alternatives to all this we could do :+1: .

server/cpho/jinja2/indicators/_progress_bar.jinja2
Stephen-ONeil: Oh, and [tooltip role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tooltip_role) is meant to go on the element containing the tooltip content. There's not a dedicated role for the element which triggers a tooltip.

Stephen-ONeil (approved):
Approving, with some known accessibility issues remaining. The upcoming accessibility review will be an opportunity to get a second opinion on the approach, and might lead to switching to an alternative to an informational tooltip.

PR #150, opened 2023-10-30
Further k8s resource optimization

Stephen-ONeil:


k8s/server/django/deployment.yaml
simardeep1792, line 1: Looks Good.

k8s/flux-system/kustomization.yaml
simardeep1792, line 1: Looks good!

simardeep1792 (commented):
Looks like a typo at the sidecar mention
    
    annotations:
      proxy.istio.io/config: '{ "holdApplicationUntilProxyStarts": true }'
      **ssidecar.istio.io/proxyCPU: 50m**
      sidecar.istio.io/proxyMemory: 112Mi

k8s/flux-system/gcr-sync.yaml
simardeep1792, line 1: If you are referring to the indentation for env and command. The indentation is correct and is the way mentioned in k8s official documentation. 
https://kubernetes.io/docs/tasks/inject-data-application/define-environment-variable-container/

k8s/server/postgres/pg-cluster.yaml
simardeep1792, line 1: Looks like a typo at the sidecar mention

annotations:
  proxy.istio.io/config: '{ "holdApplicationUntilProxyStarts": true }'
  **ssidecar.istio.io/proxyCPU: 50m**
  sidecar.istio.io/proxyMemory: 112Mi

simardeep1792:
@Stephen-ONeil Thank you for these changes. I have added some comments. Let me know if you would like to discuss further.

k8s/flux-system/gcr-sync.yaml
Stephen-ONeil: `-` delimited lists in yaml can be either indented or unintended, yaml doesn't care because in this case yaml uses the `-` as the delimiter rather than the indentation level... which is annoying, since indentation is very important in yaml in pretty much any other case, and greatly helps with readability. 

I don't know why the k8s standard seems to be _not_ indenting `-` delimited lists, but I think it's better to have the indentation. I'd originally put my `resources:` block at the wrong indentation level because of this quirk.

Stephen-ONeil:
> @Stephen-ONeil Thank you for these changes. I have added some comments. Let me know if you would like to discuss further.

@simardeep1792 thanks for catching that typo! See my response, RE: indentation of hyphen lists in yaml.

PR #151, opened 2023-11-02
Indicator home redesign.

msarar:
Removed action dropdown
All indicator related actions now in indicator homepage including: upload, export changelog and indicator edit
Column based view so that "periods with data table" receives more attention


server/cpho/jinja2/indicators/_indicator_actions.jinja2
```diff
@@ -1,4 +1,4 @@
-<div class="dropdown">
+{# <div class="dropdown">
```
Stephen-ONeil, line 1: Since this whole file is commented out, and I believe no longer used, was there a specific reason to not delete it?

server/cpho/jinja2/indicators/list_indicators.jinja2
```diff
@@ -19,12 +19,13 @@
...
           {# <th>{{ tdt('Detailed indicator') }}</th> #}
           {# <th>{{ tdt('Sub Indicator Mesaurement') }}</th> #}
-          <th>{{ tdt("Actions") }}</th>
+          <th width="15%">{{ tdt("Actions") }}</th>
         </tr>
       </thead>
       <tbody>
         {% for indicator in object_list %}
           {% if respects_rule('can_view_indicator', indicator) %}
+            {# <tr onclick="window.open('{{ url('view_indicator', args=[indicator.pk]) }}');"> #}
```
Stephen-ONeil, line 28: Similar question on this and the other commented out lines in this file, are they left around for a purpose or can they be deleted?

server/cpho/jinja2/indicators/view_indicator.jinja2
```diff
@@ -6,66 +6,96 @@
...
-        </ul>
+  <div class="container">
+    <div class="row">
+      <div class="h2">{{ tdt("Indicator Selected: ") }} {{ indicator.name }}</div>
+      <div class="col-md-2 mt-4">
+        <div>
+          {% if respects_rule('can_edit_indicator', indicator) %}
+            <a href="{{ url('edit_indicator', args=[indicator.id]) }}"
+               class="btn w-100 btn-outline-dark my-1">
+              <span class="bi bi-pen"></span>
```
Stephen-ONeil, line 18: On this and the other `<span class="bi...` elements, I recommend adding an `aria-hidden="true"` attribute. In general, for accessibility, purely decorative icons (which these are because they're redundant to the text inside the link/button) should be hidden from screen readers.

Further, for the bootstrap font icons, i.e. the `bi-...` classes, the screen readers will likely end up reading the icon's unicode (something like "F6B3") out loud by default, which would be confusing even if paired with additional text or a label sharing the same information that the icon is intended to communicate.

If the icon wasn't purely decorative (e.g. it is the whole button without accompanying text), then you would need to add an `aria-label` or equivalent with a screen reader specific explanation. In case you do run in to that case some time, heads up that an `aria-label` directly on a span will likely be ignored by many screen readers, as a span isn't expected to have meaning requiring a special explanation. You would still want to hide the span with the font icon in it, and you would put the label on the `button`/`a`/etc that the span is inside :+1:

Stephen-ONeil (changes requested):
Nice! I like these changes a lot. Couple minor comments.

AlexCLeduc (approved):
Looks great! Nothing to add over Stephen's comments

server/cpho/jinja2/indicators/view_indicator.jinja2
msarar: Thanks for taking the time to explain it. I'll surely keep this in mind. :)

server/cpho/jinja2/indicators/_indicator_actions.jinja2
msarar: I agree, we don't need to keep this file. removing it :)

server/cpho/jinja2/indicators/list_indicators.jinja2
msarar: I left this one around as this makes the entire row clickable but I wasn't sure if it meets accessibility requirements or if its best practice. (but if we ever want that feature we could comment it back in)

But I think its best to get rid of it as I don't think we will ever need it

Stephen-ONeil (approved):
:+1:

PR #156, opened 2023-11-14
Hide age group deleted data records

msarar:
age group indicator datum is hidden from user upon deletion instead of deleted completely from the system ad deleting it would delete historical data as well.

Points to note:

- new indicator datum manager: active_objects that filters out deleted objects.
- deleted objects are marked with an is_deleted bool and a deletion_time for uniqueness perspective.
- Changes made to submission status: data deletion changes local and global status to "modified since last submission".
- If data was submitted with age data, and all age group data was deleted after submission, age group will not show any data but will have status "modified since last submission".


server/cpho/jinja2/indicator_data/manage_indicator_data.jinja2
```diff
@@ -46,8 +46,9 @@
     <td>{{ form.value_displayed }}</td>
     <td>{{ form.reason_for_null }}</td>
     <td>
-      {{ form.DELETE }}
+      {# {{ form.DELETE }} #}
```
AlexCLeduc, line 49: Please remove commented out code

AlexCLeduc:
The tests are covering things excellently and make it extremely unlikely we'll break something without noticing, but I think we're still at risk of forgetting the way deletion works in a year. Could we have explicit checks for the following:
- Deleting via manual input sets `is_deleted` and `deleted_time` (no need to check time is recent) on the record? 
- Submitting data that includes deleted records does in fact set prog/hso_submitted=True on versions with is_deleted=True
    - Document this assertion/block with a comment - relying on submitted deleted versions was a decision and we don't want anyone to think it was an accident

server/tests/test_soft_delete.py
```diff
@@ -0,0 +1,194 @@
...
+        "global_status": SUBMISSION_STATUSES.PROGRAM_SUBMITTED,
+    }
+
+    url = reverse("submit_indicator_data_all", args=[indicator_id, period.id])
+    with patch_rules(can_submit_as_hso_or_program=True):
+        resp = vanilla_user_client.post(url, {"submission_type": "hso"})
+        assert resp.status_code == 302
+
+    statuses = get_submission_statuses(indicator, period)
+
```
AlexCLeduc, line 109: This could be a good place to assert the version has `is_deleted=True, program_submitted=True, hso_submitted=True`

server/tests/test_manual_data_input.py
```diff
@@ -216,7 +216,7 @@ def test_agegroups_existing_data(vanilla_user_client):
         response = vanilla_user_client.post(url, data=data)
         assert response.status_code == 403
     ind.refresh_from_db()
-    assert ind.data.filter(period=period).count() == 4
+    assert ind.data.filter(period=period, is_deleted=False).count() == 4
```
AlexCLeduc, line 219: This is a good place for an explicit check a refreshed `record_51_75` has `is_deleted` and `deleted_time`

AlexCLeduc (approved):

PR #158, opened 2023-11-16
Active directory SSO

Stephen-ONeil:
As implemented in ITAP.


Stephen-ONeil:
Will have to "test in prod" on this one. The SSO endpoint can't be tested locally as it checks for the prod host name, and I can't deploy this before merging as flux's auto-reconciliation will force the cluster state back to whatever's on the main branch.

server/cpho/jinja2/login.jinja2
```diff
@@ -2,63 +2,79 @@
...
-                     class="textinput textInput form-control">
-              <label for="{{ form.password.id_for_label }}">{{ form.password.label_tag() }}</label>
+            <p>{{ tdt("Welcome") }}</p>
+            <p>{{ tm("prod_login_intro_2") }}</p>
+            {#
+              <p>
+                {{ tm("prod_login_intro_3") }}
+                <a href="{{ url("user_agreement") }}">{{ tm("user_agreement") | lower }}</a>.
+              </p>
+            #}
```
Stephen-ONeil, line 26: Could delete this, and the corresponding translation entry, if you don't expect to need a user agreement on this project. Don't know if ITAP has it because the client requested it, or if it was required during approvals (and therefore is needed here too).

AlexCLeduc: Go ahead and delete it, let's try and slip by without it, ITAP was special

server/server/middleware.py
```diff
@@ -13,12 +13,18 @@ def __call__(self, request):
         return self.get_response(request)
 
     def process_view(self, request, view_func, view_args, view_kwargs):
-        if getattr(view_func, "allow_unauthenticated", False):
+        third_party_allow_unauthenticated_url_names = [
+            "phac_aspc_helper_login"
```
AlexCLeduc, line 17: do we also need `phac_aspc_helper_authorize` ? I'm honestly not sure but we whitelisted both in cedars

AlexCLeduc (approved):
Looks great, thanks!

server/server/middleware.py
Stephen-ONeil: Ah, yup, looks like it would be necessary, good catch!

PR #165, opened 2023-11-29
added form validation for benchmarking

kdompaul:
added form validaiton for the new fields value, standard deviation, and year. Did not add add form validation for labels as it will be switching to a dropdown menu

### DIFF
--- a/server/cpho/jinja2/benchmarking/manage_benchmarking_data.jinja2
+++ b/server/cpho/jinja2/benchmarking/manage_benchmarking_data.jinja2
@@ -11,9 +11,30 @@
 {% macro benchmarking_form(form) %}
   <tr class="benchmarking-form">
     <td>{{ form.oecd_country }}</td>
-    <td>{{ form.value }}</td>
-    <td>{{ form.year }}</td>
-    <td>{{ form.standard_deviation }}</td>
+    <td>
+      {{ form.value }}
+      {% if form.errors.value %}
+        <div class="error-display">
+          {% for error in form.value.errors %}<strong>{{ error|escape }}</strong>{% endfor %}
+        </div>
+      {% endif %}
+    </td>
+    <<td>
+      {{ form.year }}
+      {% if form.errors.year %}
+        <div class="error-display">
+          {% for error in form.year.errors %}<strong>{{ error|escape }}</strong>{% endfor %}
+        </div>
+      {% endif %}
+    </td>
+    <<td>
+      {{ form.standard_deviation }}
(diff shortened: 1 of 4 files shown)


server/cpho/views/indicator_metadata.py
```diff
@@ -109,6 +107,54 @@ class Meta:
         ),
     )
 
+    def clean(self):
+        cleaned_data = super().clean()
+        return cleaned_data
```
msarar, line 112: I don't think we need this function because all we are doing is calling the super.clean() function.

msarar:
@kdompaul Tests are failing; will need to update the test for benchmarking as more fields are required for form submission now

server/cpho/jinja2/benchmarking/manage_benchmarking_data.jinja2
```diff
@@ -10,11 +10,46 @@
...
-    <td>{{ form.year }}</td>
-    <td>{{ form.standard_deviation }}</td>
-    <td>{{ form.comparison_to_oecd_avg }}</td>
+    <td>
+      {{ form.oecd_country }}
+      {% if form.errors.oecd_country %}
+        <div class="error-display">
+          {% for error in form.oecd_country.errors %}<strong>{{ error|escape }}</strong>{% endfor %}
+        </div>
+      {% endif %}
```
AlexCLeduc, line 19: This is often repeated, can we put the entire block including the conditional in its own macro? e.g. `{{ benchmarking_form_errors(form.oecd_country) }}` it can be defined above this macro

server/cpho/views/indicator_metadata.py
```diff
@@ -109,6 +107,67 @@ class Meta:
...
+        oecd_country = self.cleaned_data["oecd_country"]
+        if not oecd_country:
+            self.add_error("oecd_country", tdt("Please select a country"))
+
+        return oecd_country
+
+    def clean_value(self):
+        value = self.cleaned_data["value"]
+        if value and value < 0:
+            print("invalid")
```
AlexCLeduc, line 130: please remove the print statement 😄

server/static/cpho.css
```diff
@@ -52,6 +52,11 @@ td:has(>.error-display)>input {
 
 }
 
+td:has(>.error-display)>select {
+    border: 2px solid #75070c;
+
+}
+
```
AlexCLeduc, line 59: I forgot about `:has`, this is really simple and looks great!

AlexCLeduc (approved):
Thanks!

PR #170, opened 2023-12-08
Make a11y-testing, temporarily, the image deployment and cluster syncing branch

Stephen-ONeil:


k8s/flux-system/gotk-sync.yaml
```diff
@@ -8,7 +8,7 @@ metadata:
 spec:
   interval: 1m0s
   ref:
-    branch: "main"
+    branch: "a11y-testing"
```
Stephen-ONeil, line 11: This file starts with the comment `# This manifest was generated by flux. DO NOT EDIT.`, I believe because the expected update procedure is to re-run `flux bootstrap` with the necessary changes. Except I don't know what the precise set of arguments provided to the original `flux bootstrap` call were, and would potentially make other inadvertent changes while trying to guess!

On the other hand, I believe that updating this one line in the definition file is safe. When this merges to main, flux should sync it up fine and then switch to watching the `a11y-testing` branch. ...although I guess that means I should make sure these changes are merged in to the `a11y-testing` branch first, before this merges to main, in case that causes a feedback loop, ha.

Am I missing something here @vedantthapa? If you think you know the original `flux bootstrap` arguments or have reason to believe updating the sync branch this way could cause problems, chime in!

vedantthapa (approved):
Fire away!

PR #172, opened 2023-12-12
Configure SOPS

vedantthapa:
Configure SOPS in favour of SealedSecrets.

[How it works with flux?](https://fluxcd.io/flux/guides/mozilla-sops/)
[How do I encrypt my secrets?](https://github.com/getsops/sops/blob/main/README.rst#23encrypting-using-gcp-kms)

Only encrypt the `data` / `stringData` field for the k8s secret and not the entire yaml file. A base configuration is provided in the `.sops.yaml` file, therefore, the command to encrypt secret is `sops -e <path-to-k8s-secret> > <path-to-write-output>` from the root of the repo.

The SealedSecrets controller will be deprecated in a separate PR.


vedantthapa:
There was some additional configuration pushed with https://github.com/PHACDataHub/cpho-phase2/commit/2743e3ffc79d7d79e8cd3e13bb4c533116ab6e45 to make this work with flux.

In general, to use KMS encrypted secrets we'd need to add the `decryption` block i.e, -
```yaml
...
  decryption:
    provider: sops
``` 
to a flux `kustomization` resource in the `sync.yaml` (see [this](https://github.com/PHACDataHub/cpho-phase2/blob/main/k8s/server/sync.yaml#L33-L48), for full example). Currently only the `flux-system` and `server-django` kustomization resources have it.

PR #173, opened 2023-12-15
Update sync manifests

vedantthapa:
tldr;
- add new `gitrepo` resource for `a11y-testing`
- update `server` namespace to reconcile against the new `gitrepo` resource
- point all other namespaces back to reconciling against the `main` branch

In the current scenario, all resources are reconciled against the `a11y-testing` branch. This makes it difficult to make  changes to the kubernetes specific components cause then we'd have to merge to `main` and then to `a11y-testing` to keep things clean.

This PR addresses the problem by adding a new `gitrepo` resource pointing to `a11y-testing` and updating the sync manifest in the `server/` directory to reconcile against the new `gitrepo` resource. In other words, this means all resources other than the ones in `server` namespace reconcile back to the `main` branch and all resources in `server` namespace reconcile to the `a11y-testing` branch.

### DIFF
--- a/k8s/flux-system/gotk-sync.yaml
+++ b/k8s/flux-system/gotk-sync.yaml
@@ -8,7 +8,7 @@ metadata:
 spec:
   interval: 1m0s
   ref:
-    branch: "a11y-testing"
+    branch: "main"
   secretRef:
     name: flux-system
   url: ssh://git@github.com/PHACDataHub/cpho-phase2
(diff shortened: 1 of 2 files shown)


Stephen-ONeil (approved):
Approved, with the caveat that we don't do anything that might cause any significant downtime for the next month or so, while a11y testing is ongoing. Doesn't really need to be said, but we'll also need to be careful not to change anything in the infrastructure that would require a corresponding app code change (not common, but e.g. we can't change the cluster IP since the app needs to know it).

Issue #159, opened 2023-11-21
Dependent pods should restart when SealedSecret resources are modified.


Stephen-ONeil:
Updated one of our SealedSecret-stored app config values, had to manually restart the app pod to get the new value in to its env vars. Not at all ideal, should be automatic, with rolling updates to avoid down time and all that nice-ness.

Stephen-ONeil:
Update: or perhaps it would have if the secret was _actually_ successfully synced, as the flux status notifications claimed
![image](https://github.com/PHACDataHub/cpho-phase2/assets/11301919/5ebcd81a-7ef3-4cc3-9718-372ae8147c79)

Stephen-ONeil:
Fixed the above update, unrelated bug, the application pod still required a restart in the end.

vedantthapa:
I believe this is more of a [kubernetes thing](https://kubernetes.io/docs/tasks/inject-data-application/distribute-credentials-secure/#define-container-environment-variables-using-secret-data) rather than a SealedSecret issue -
> If a container already consumes a Secret in an environment variable, a Secret update will not be seen by the container unless it is restarted. There are third party solutions for triggering restarts when secrets change.

That is to say, it'd still exist when we migrate to SOPS unless we use a third-party solution like https://github.com/stakater/Reloader. However, that implies more pods.

PR #175, opened 2023-12-19
Update SSO Secret

vedantthapa:
As mentioned in https://github.com/PHACDataHub/cpho-phase2/issues/159#issuecomment-1852400501, the deployment might require a restart to propagate the updated secret value.

Also, this branch needs to be further merged into `a11y-testing` due to https://github.com/PHACDataHub/cpho-phase2/pull/173.

### DIFF
--- a/k8s/server/django/secrets.enc.yaml
+++ b/k8s/server/django/secrets.enc.yaml
@@ -15,8 +15,8 @@ sops:
     azure_kv: []
     hc_vault: []
     age: []
-    lastmodified: "2023-12-12T18:45:36Z"
-    mac: ENC[AES256_GCM,data:wHnK4XuiObfZG7CK2+bAZIIb2REYfpwluLn7hVwJx6+rtLb69H1aELAJpdD5WQAqB+7JhMjUPWBDsUYHBO7kkoL+PMSYDlJ0R70p+rHfCQgo2SpBBAh5BvYRuwmYH6hpQ+wsOQ7hvn2RSNpKUtQyYoSh2lThsF0IFDB1+Iobbh4=,iv:KMEem2wwdrE/+DOkE+uiz6b2B/IDImiR89XYGJa85/g=,tag:Sdq4HawBj/PLfJzVliZDCg==,type:str]
+    lastmodified: "2023-12-19T21:03:59Z"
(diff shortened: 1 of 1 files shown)


Stephen-ONeil (approved):
I say go for it, though I was late to the discussion yesterday so I may have missed some of the reasons to delay.

AlexCLeduc (approved):
Kind of confusing what's changed, why are all all the lastModified's updated, aren't we only really changing the oauth config? And is that a single value or several key/value pairs?

AlexCLeduc:
Please make sure the "old" plain auth is still working, we don't want DISABLE_LEGACY_LOGIN, but I can't really tell if that's part of of this or not

Stephen-ONeil:
> Kind of confusing what's changed, why are all all the lastModified's updated, aren't we only really changing the oauth config? And is that a single value or several key/value pairs?

The three oAuth secrets are all key-value items inside a single k8s secret resource. I believe updating any value in the secret requires re-encrypting all values, with SOPS. @vedantthapa can you confirm that's what's happening here?

> Please make sure the "old" plain auth is still working, we don't want to turn off ENABLE_LEGACY_LOGIN, but I can't really tell if that's part of of this or not

`ENABLE_LEGACY_LOG_IN` is set in plain text in `k8s/server/django/deployment.yaml`, not changed here, so all good on that :+1:

vedantthapa:
> Kind of confusing what's changed, why are all all the lastModified's updated, aren't we only really changing the oauth config?

I agree it's a bit confusing. I believe the `last-modified` fields were updated because I used the sops editor to edit the file. There's just one file that has all the secrets - `k8s/server/django/secrets.enc.yaml`

> And is that a single value or several key/value pairs?

Only a single key value pair, i.e, `client-secret` field in the `phac-aspc-oauth-config` secret was updated. Diff on L27 - https://github.com/PHACDataHub/cpho-phase2/pull/175/files#diff-8db078c1afb6283c37fbe18bc1aef6c80599175bd5dcb6e5809bb2b21abb8da8R27

vedantthapa:
If it helps, we can organize individual secrets in their own files, i.e, one file per secret? That might clean up the diff when updating a secret value.

However, that also implies - anytime a new secret is created, it must be added to the `kustomization.yaml`

simardeep1792 (approved):

AlexCLeduc:
> If it helps, we can organize individual secrets in their own files, i.e, one file per secret? That might clean up the diff when updating a secret value.
> 
> However, that also implies - anytime a new secret is created, it must be added to the `kustomization.yaml`

I was just curious, no need to complicate or change any processes.  Thanks for the explanation

PR #180, opened 2024-01-03
Simsingh/cert manager

simardeep1792:
Changes:
- Removed the Chainguard Images for the cert-manager
- Updated the CRDs for the cert-manager
- Removed certificate.yaml from the cert-manager folder
- Refactored the let's encrypt certificate from Issuer to Cluster Issuer


k8s/cert-manager/kustomization.yaml
```diff
@@ -2,4 +2,29 @@ apiVersion: kustomize.config.k8s.io/v1beta1
...
 resources:
   - issuers.yaml
-  - certificate.yaml
+patches:     
+- target:
+    group: cert-manager.io
+    version: v1
+    kind: ClusterIssuer
+    name: letsencrypt
+    namespace: istio-system
```
vedantthapa, line 11: Since it's a `ClusterIssuer`, I don't think we'll be needing this.

vedantthapa (changes requested):

k8s/cert-manager/kustomization.yaml
simardeep1792: Good catch! fixed and pushed

Stephen-ONeil (approved):
:+1:

vedantthapa:
@simardeep1792 if things are all set here can we merge this?

Issue #183, opened 2024-01-10
Path based vs sub domain based routing


vedantthapa:
In order to expose a deployment in an ephemeral environment, we can use [istio virtual service](https://istio.io/latest/docs/reference/config/networking/virtual-service/#VirtualService) with either _path_ based routing i.e, `/prod` endpoint pointing to prod instance and `/ephemeral/pr-123` pointing to an ephemeral instance that is created off of a PR **OR** _sub-domain_ based routing i.e, `prod.hopic-sdpac.phac-aspc.alpha.canada.ca` pointing to prod instance and `*.dev.hopic-sdpac.phac-aspc.alpha.canada.ca` pointing to ephemeral instances where `*` can be a PR id.

Path based routing has it's benefits in terms of SSO because the redirect URIs only need to be configured once, however, it might require some changes to the application code. Sub-domain based routing, on the other hand, does not require app updates, but also does work with SSO unless we can get a wildcard redirect URI.

vedantthapa:
Path based routing doesn't deal with redirects well. It would require the application to expose a route prefix env variable which can then be configured with istio virtual serivce's [`HTTPRoute`](https://istio.io/latest/docs/reference/config/networking/virtual-service/#HTTPRoute).

A long term solution would be have SSO as an infrastructure component with either [Google IAP](https://cloud.google.com/security/products/iap?hl=en) / [OAuth2 proxy](https://oauth2-proxy.github.io/oauth2-proxy/).

re: Disabling SSO for ephemeral environments
> Disabling SSO opens a hole in the app's security boundary, since they rely on SSO as an alternative to being PHAC VPN locked. Definitely a no-go if/when the devs want to use copies of prod data in the ephemeral environment (maintaining realistic fake data for testing is a burden we can't enforce)

vedantthapa:
After merging #185 i.e, exposing a route prefix env variable `SCRIPT_NAME`, stuff works as expected. It's important to note that the `SCRIPT_NAME` env variable in the deployment must have a leading `/` (example: `/pr1`, `/pr2`, etc.) otherwise it'll mess up redirections after the initial login.

[...]

PR #185, opened 2024-01-16
Support SCRIPT_NAME env var, to enable "ephemeral" dev builds to be hosted from unique paths under a shared subdomain 

Stephen-ONeil:
Not super well documented on Django's end, and a little wonky, but this is a common feature of many servers/server gateway interfaces and it's Django's prescribed method for hosting a Django app from a sub-resource (path) under a domain. I've left a beefy comment in settings.py to cover the weirder aspects. The env var will only be set in the ephemeral deployments, so none of this impacts prod or local dev.

Well, one caveat to that. Code that modifies the path directly (notably and possibly only `convert_url_other_lang`) needs to understand and account for `SCRIPT_NAME` path components, or else there will be ephemeral deployment specific bugs. Not a big deal though, as I see it. Forgetting to account for it won't impact prod, worst case it's just an obscure foot gun for the obscure ephemeral dev builds.

### DIFF
--- a/server/cpho/jinja_helpers.py
+++ b/server/cpho/jinja_helpers.py
@@ -1,3 +1,4 @@
+import os
 from urllib.parse import quote, urlencode, urlparse, urlunparse
 
 from django.conf import settings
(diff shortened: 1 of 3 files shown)


vedantthapa (approved):
Nice! The code comment is super helpful in understanding the behaviour. One question though - is `STATIC_NAME` env variable (in title) different from `SCRIPT_NAME` env variable (in code comments)? / is this a typo?

Stephen-ONeil:
> Nice! The code comment is super helpful in understanding the behaviour. One question though - is `STATIC_NAME` env variable (in title) different from `SCRIPT_NAME` env variable (in code comments)? / is this a typo?

Oops, that was a typo! Meant `SCRIPT_NAME`, thanks for catching that, haha.

AlexCLeduc (approved):

vedantthapa:
Relates to #183

PR #189, opened 2024-01-24
Indicator directory authorization

AlexCLeduc:
Replaces phac-org with indicator-directories: just a simple name/description model with m2m fields towards users and indicators. 

User management now looks like this:
Screenshot 2024-01-24 at 16 39 08

directory form looks like:
- it also has a link to the plain user's page, so curious admins from HSO can see this page without switching logins
Screenshot 2024-01-24 at 16 39 32

logging in as a user who is in one indicator-group:
Screenshot 2024-01-24 at 16 40 29

users in a group can add new people to that group, via this page (linked to from above screenshot)
Screenshot 2024-01-24 at 16 41 24


server/cpho/models/metadata.py
```diff
@@ -31,11 +31,69 @@ class IndicatorDataSubmission(models.Model):
...
-        "cpho.User", on_delete=models.CASCADE, related_name="phac_org_roles"
+        "cpho.User",
+        on_delete=models.CASCADE,
+        related_name="indicator_directory_user_accesses",
     )
-    phac_org = fields.ForeignKey(
-        "cpho.PHACOrg", on_delete=models.CASCADE, related_name="phac_org_roles"
+    directory = fields.ForeignKey(
+        "cpho.IndicatorDirectory",
+        on_delete=models.CASCADE,
```
AlexCLeduc, line 91: I'm trying something a bit weird here: I'm defining explicit through models for the m2m relations. This allows us to be flexible and add m2m fields on the user or indicator models, making it extra-easy to define forms. 

Note that the directory isn't a versioned model, so I'm not sure if versioning works properly here. If we ever try explicit through-models on a versioned model, we'll want explicit tests to check everything is versioning properly.

msarar (approved):
Neat!
Thanks Alex

PR #191, opened 2024-01-26
5368 Indicator Metadata

msarar:
First pass at adding indicator metadata fields to the indicator model.
Added quintiles
Added indicator fixtures (json and yaml)

HSO is currently working on providing (google sheet, column) pairs that are necessary 
If we have a some of them implemented, it'll better help them visualize and and create requirements.


AlexCLeduc (approved):

AlexCLeduc:
We have a lot of repeated markup in the indicator form. A macro would be a good idea. Maybe we can even offload the rendering to the `_standard_form.jinja2` template (although we'd have to add the mark_safe in there too, and potentially list an explicit field order in the form class)

AlexCLeduc (approved):

PR #194, opened 2024-02-07
Configure flux for noble space

vedantthapa:
- Update project id at relevant places
- Re-encrypt secrets

### DIFF
--- a/.sops.yaml
+++ b/.sops.yaml
@@ -2,4 +2,4 @@ creation_rules:
   # for all yaml files, encrypt data / stringData fields using GCP KMS
   - path_regex: .*.yaml
     encrypted_regex: '^(data|stringData)$'
-    gcp_kms: projects/pht-01hhjj7tt0m/locations/northamerica-northeast1/keyRings/sops/cryptoKeys/sops-flux-key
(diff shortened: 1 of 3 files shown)


Stephen-ONeil (approved):
LGTM.

Is the flux slack notification configuration handled in here by the way? I don't think I ever asked how that worked. On my mind because the (non-k8s) uptime monitoring checks and their slack configuration isn't in IaD yet. I'll use the old pulumi to bring them back for now, but it's something to port at some point (and generating a slack token etc will need to be documented as a manual step for bootstrapping).

vedantthapa:
> Is the flux slack notification configuration handled in here by the way? I don't think I ever asked how that worked. On my mind because the (non-k8s) uptime monitoring checks and their slack configuration isn't in IaD yet. I'll use the old pulumi to bring them back for now, but it's something to port at some point (and generating a slack token etc will need to be documented as a manual step for bootstrapping).

Yep, flux slack notifications are handled here. Essentially, we just used the slack token that you set up for non-k8s uptime monitoring. Here's some info on flux alerts: https://fluxcd.io/flux/monitoring/alerts/.

PR #195, opened 2024-02-08
Configure server for noble

vedantthapa:
- Update `cnpg-system` CRDs
- Update project specific dependencies - project-ids, service account names and re-encrypt secrets.
- Add a `ConfigMap` resource for storing configurations. This is referenced in the **overlay** deployments (not in base) via a `envFrom` key
- Update resources for server pod.
   Since the minimum resources for the server pod was causing a scale-up by the HPA in our old cluster, I've bumped up the resources a bit to ensure there's only one replica. The HPA would still "kick-in" if there's a spike in traffic.
- Update resources for postgres cluster pods
   Don't exactly remember why but the postgres cluster pods were using `1926Mi` memory. Based on the utilization from the old cluster, I've dialed down the cpu / memory on this.
- Move resources for `istio-proxy` container to base deployment spec.

Depends on #196.

### DIFF
--- a/k8s/cnpg-system/cnpg-cm.yaml
+++ b/k8s/cnpg-system/cnpg-cm.yaml
@@ -4,5 +4,5 @@ metadata:
   name: cnpg-controller-manager-config
   namespace: cnpg-system
 data:
-  INHERITED_LABELS: app
+  INHERITED_LABELS: app, branch
   INHERITED_ANNOTATIONS: proxy.istio.io/*, sidecar.istio.io/*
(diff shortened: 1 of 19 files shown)


Stephen-ONeil (approved):
Nice! LGTM!

>Update resources for server pod.
Since the minimum resources for the server pod was causing a scale-up by the HPA in our old cluster, I've bumped up the resources a bit to ensure there's only one replica. The HPA would still "kick-in" if there's a spike in traffic.

Nice, hm, if we aren't using minimum pod sizes then I'll re-enable gunicorn process forking to get more out of the resources. Never really planned to leave forking disabled, Mike just wanted us to see whether we could operate with one server process per pod but I think it's clear that's inefficient. 

>Update resources for postgres cluster pods
Don't exactly remember why but the postgres cluster pods were using 1926Mi memory. Based on the utilization from the old cluster, I've dialed down the cpu / memory on this.

I think I set it high as an initial guess, never went back to look at the usage and adjust. In part, that's because we've not been operating under realistic load/haven't done any load testing. Dialing it down for now is a good idea since it'll a) save a little money and b) give us more data on resource usage to performance. If we don't work out load testing before launch, we'll just have to watch it in prod and further right-size/set scaling as needed.

vedantthapa:
Bootstrapping the DB from the old cluster didn't require copying the backup to the new storage bucket. Instead, I assigned IAM permissions to the `hopic-postgres-backup` service account to read data from the old project's bucket during the recovery process.

PR #196, opened 2024-02-08
Configure ASM and cert-manager for noble

vedantthapa:
- Update project specific dependencies - project-ids, service account names and load balancer ip.
- Update dependency order
   Initially `cert-manager` used to depend on the `istio-ingress` ks, however, this has now been reversed due to fact that we expect the certificate secret to be created before gateway configuration can consume it.
- Clean up comments

Depends on https://github.com/PHACDataHub/dns/pull/56 to work.


simardeep1792 (approved):
LGTM!

Stephen-ONeil (approved):
:+1:

PR #197, opened 2024-02-08
Configure infrastructure for NIST compliance

vedantthapa:
As per the recommendations for NIST compliance in the GCP's security command center -
- Set key rotation period to 60 days
  As per https://github.com/getsops/sops/issues/1135#issuecomment-1731811794, it shouldn't cause any issues to existing secrets when a key is rotated.
- Enable flow logs on subnet
  I "clickops-ed" this because it's not a part of IaD, however, added the relevant flag to `k8s/Taskfile` command for future use.
- Enable IAM Audit logs

In addition to this, the `infrastructure/` directory is added to the cloudbuild trigger's ignored files.

### DIFF
--- a/infrastructure/cloudbuild.yaml
+++ b/infrastructure/cloudbuild.yaml
@@ -16,3 +16,4 @@ spec:
       branch: .*
   ignoredFiles:
     - "k8s/**"
+    - "infrastructure/**"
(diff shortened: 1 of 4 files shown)


Stephen-ONeil (approved):
:+1: :+1:

PR #201, opened 2024-02-09
Add `allowPrivilegeEscalation` to deployments

vedantthapa:
Even after merging https://github.com/PHACDataHub/cpho-phase2/pull/199, the security posture dashboard tagged `cnpg-controller-manager` and `server` deployments as misconfigurations under `pod container allows privilege escalations on exec`.

This was due to missing `allowPrivilegeEscalation` field for the `istio-proxy` container in the deployment spec (thanks to @simardeep1792 for pointing this). The alert technically is a false positive because the pods created by these deployments still have the `allowPrivilegeEscalation` field set. 

In any case, this PR explicitly sets the security context for `istio-proxy` containers to clean up the dashboard.

### DIFF
--- a/k8s/cnpg-system/kustomization.yaml
+++ b/k8s/cnpg-system/kustomization.yaml
@@ -35,6 +35,8 @@ patches:
                   requests:
                     cpu: 50m
                     memory: 112Mi   
+                securityContext:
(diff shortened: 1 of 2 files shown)


simardeep1792 (approved):
LGTM!

PR #204, opened 2024-02-09
Add ephemeral env setup

vedantthapa:
- Configure subdomains in infrastructure with `*.dev.hopic-sdpac.phac-aspc.alpha.canada.ca`. The `*` would be replaced by a branch name in the actual deployment.
- Update certificate to include the new domain (wildcard cert).
- Add `./k8s/server/ephemeral-instances/` dir where all the flux syncs related to ephemeral builds will be stored.
  I've added a config for you @Stephen-ONeil with your name. We'll probably change this to follow `<BRANCH_NAME>-sync` convention when your branch is ready. Also need to replace `${BRANCH_NAME}` with an actual branch name in that file.
- Update DB configuration to bootstrapped a cluster from the cloud storage bucket.
  This should be replaced in the future to use dev seed scripts.


k8s/server/overlays/ephemeral/postgres/pg-cluster.yaml
```diff
@@ -1,20 +1,40 @@
...
 metadata:
   name: cpho-postgres14-cluster
 spec:
   instances: 1
-  resources:
-    requests:
-      cpu: 450m
-      memory: 1926Mi
+  startDelay: 900
+  stopDelay: 900
```
Stephen-ONeil, line 9: Just curious, what are the delay's for?

Stephen-ONeil (approved):
Thanks!

> I've added a config for you @Stephen-ONeil with your name. We'll probably change this to follow <BRANCH_NAME>-sync convention when your branch is ready. Also need to replace ${BRANCH_NAME} with an actual branch name in that file.

Branch pushed as `fix-cloud-trace-exporting`, should have an image built soon.

> Update DB configuration to bootstrapped a cluster from the cloud storage bucket.
This should be replaced in the future to use dev seed scripts.

I think the seed scripts would generally be better too, but it'll be the devs call and they might prefer one or the other depending on needs. E.g. general testing might be fine with a freshly seeded DB/crafted fake data, but they may also want to use ephemeral builds to debug a specific data issue seen in prod some time.

Stephen-ONeil:
FYI: initial branch image built as `fix-cloud-trace-exporting-3e9f421b-1707502348`

k8s/server/overlays/ephemeral/postgres/pg-cluster.yaml
vedantthapa: The `stopDelay` is to give a postgres cluster pod time to complete it's operations (mostly related to WAL) in the event of an unexpected / expected pod deletion.

The `startDelay` is to delay probe executions (liveliness / readiness). I think this was configured due to fact that `istio-injection` sometimes takes a while to completed.

Here's the doc reference: https://cloudnative-pg.io/documentation/1.20/instance_manager/#startup-liveness-and-readiness-probes

vedantthapa:
> FYI: initial branch image built as `fix-cloud-trace-exporting-3e9f421b-1707502348`

Cool! Once this is merged into main, you'll need to merge main into your branch so that the updated k8s configurations are reflected.

PR #207, opened 2024-02-13
Add K8s and cloud permissions for server deployment

vedantthapa:
- Add a GCP service account for server deployment
  Any permissions for the prod server pod must be assigned to this GCP service account.
- Add `roles/cloudtrace.agent` to fix cloud trace permission issues on server pod.
- Add K8s service account resource with appropriate annotations for workload identity and attach it to `server` deployment.
  Note that the GCP service account annotation is only added to the **prod** server deployment and not the **ephemeral** one due to issues with binding workload identity user permission ephemerally as highlighted in https://github.com/PHACDataHub/cpho-phase2/pull/205#issue-2130770901. Ephemeral deployment pods will use the `server` K8s service account from their respective namespaces but, unlike prod, won't have any permissions on the cloud.

### DIFF
--- a/infrastructure/iam-mapping.yaml
+++ b/infrastructure/iam-mapping.yaml
@@ -9,6 +9,11 @@ spec:
     - members:
         - memberFrom:
             serviceAccountRef:
+              name: server-deployment
+      role: roles/cloudtrace.agent
+    - members:
+        - memberFrom:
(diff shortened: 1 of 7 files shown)


Stephen-ONeil (approved):
:+1:

simardeep1792 (approved):

PR #208, opened 2024-02-13
Add alert message to exclusion list

vedantthapa:
Flux is spamming `Namespace/cnrm-system configured` on the slack updates channel and I can't seem to figure out the exact reason.

As a workaround, add the specific message to `exclusionList` key.


Stephen-ONeil (changes requested):
Can you add a brief comment in the file for why it needed to be excluded, and why it's safe to exclude? Would be great if you could do that for the dependencies item too.

Stephen-ONeil (approved):
Thanks!

PR #212, opened 2024-02-14
Recreate K8s cluster

vedantthapa:
- Update Taskfile to include steps for re-creating the GKE cluster with `gke-default` service account.
- Update cluster name in flux alerts.
- Update DNS from `*.dev` to `*.eph` for ephemeral builds.
   This is mainly to accommodate branch names like `dev-test`.
- Update branch label to `prod` in `./k8s/server/overlays/prod` deployments.
- Add `roles/artifactregistry.reader` role for `gke-default` service account to be able to read images from the artifact registry.

This branch probably needs to be merged into both `main` and `prod`.

**Going forward all changes for `hopic-sdpac.phac-aspc.alpha.canada.ca` will be deployed from `prod` branch.**


vedantthapa:
I'll deploy the `server` once this branch is merged into both `main` and `prod`.

Stephen-ONeil (approved):
Approved, although personally I prefer `.dev.` to `.eph.`, I think it's clearer what it means at a glance. We know "eph" is short for ephemeral because we know that the temporary/dev builds are referred to as "ephemeral" internally, but if we send the link to someone outside the project then they aren't going to be able to guess that. They'd probably understand that "dev" is short for development though.

vedantthapa:
> Approved, although personally I prefer `.dev.` to `.eph.`, I think it's clearer what it means at a glance. We know "eph" is short for ephemeral because we know that the temporary/dev builds are referred to as "ephemeral" internally, but if we send the link to someone outside the project then they aren't going to be able to guess that. They'd probably understand that "dev" is short for development though.

On a second thought, I agree to this. As mentioned in the PR description, it was only to accommodate dev-test cause `dev-test.dev.hopic-sdpac.phac-aspc.alpha.canada.ca` didn't really sound right. I've reverted back to `dev`.

PR #214, opened 2024-02-14
Add ephemeral build for `dev-test`

vedantthapa:
Add ephemeral build for main branch. 

The endpoint for this deployment would be `main.dev.hopic-sdpac.phac-aspc.alpha.canada.ca` due to the fact that resources for ephemeral builds are modified based on the `${BRANCH_NAME}` variable and therefore the virtual service configuration ends up being - https://github.com/PHACDataHub/cpho-phase2/blob/be228da5c53fbe65ba4cdf9f72becb31d90f23c5/k8s/server/overlays/ephemeral/django/virtual-service.yaml#L7

Having `dev-test` would break this convention. Unless, we have a separate branch for `dev-test` and spin up an env based off of it.

### DIFF
--- a/k8s/server/ephemeral-instances/main-sync.yaml
+++ b/k8s/server/ephemeral-instances/main-sync.yaml
@@ -1,0 +1,110 @@
+apiVersion: source.toolkit.fluxcd.io/v1
+kind: GitRepository
+metadata:
+  name: main
+  namespace: flux-system
+spec:
+  interval: 1m0s
+  ref:
+    branch: main
+  secretRef:
+    name: flux-system
+  url: ssh://git@github.com/PHACDataHub/cpho-phase2
+---
+apiVersion: kustomize.toolkit.fluxcd.io/v1
+kind: Kustomization
+metadata:
+  name: main-infra
+  namespace: flux-system
+spec:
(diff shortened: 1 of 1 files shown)


Stephen-ONeil (changes requested):
Confirm with Alex, but I believe what they want is a separate `dev-test` branch with an ephemeral environment, and probably no ephemeral environment for `main`. `main` is for staging code before releasing it by pushing to prod, and they might want to use `dev-test` for things that aren't ready to stage yet.

vedantthapa:
> Confirm with Alex, but I believe what they want is a separate `dev-test` branch with an ephemeral environment, and probably no ephemeral environment for `main`. `main` is for staging code before releasing it by pushing to prod, and they might want to use `dev-test` for things that aren't ready to stage yet.

Ah, my bad!

@AlexCLeduc Before I make the necessary changes to this PR, could you please confirm this?

AlexCLeduc:
Yes that is correct, 2 envs, one for prod and an ephemeral one for dev-test

Stephen-ONeil (approved):
:+1:

vedantthapa:
Waiting for https://github.com/PHACDataHub/cpho-phase2/pull/216 to merge.

PR #215, opened 2024-02-15
extend timeout to 3 days

AlexCLeduc:
I just want to add a non-secret environment variable, is this the correct way to do it? Thanks!


vedantthapa (approved):
Yep!

PR #217, opened 2024-02-15
Fix unauthenticated redirects from /healthcheck route

Stephen-ONeil:
Pretty trivial. Health checks were triggering the unathenticated middleware and getting redirected to the login page. Marking the /healthcheck route as allowing unathenticated traffic prevents that, which reduces the general logging and telemetry noise from healthchecks as well as their impact on server resources.


vedantthapa (approved):
LGTM, thanks!

PR #219, opened 2024-02-19
Add DNS zone for ops

vedantthapa:
Add DNS zone for operations hopic endpoint i.e, `hopic-sdpac.data-donnes.phac-aspc.gc.ca`.

This is mostly to get SSO redirect set up with the new URL and test if DNS delegation works in the [upstream project](https://github.com/PHACDataHub/phac-dns).


Stephen-ONeil (approved):
:+1:

PR #221, opened 2024-02-20
Add sync for `infrastructure/`

vedantthapa:
- Add sync for `infrastructure/`
- Update name for cloud build trigger from `cloudbuild.yaml` to `cloudbuildtrigger.yaml`.
   For some reason flux was running into the following error while executing it's `kustomize build` command to build up the resources.  
   ```
  kustomize build failed: accumulating resources: accumulation err='accumulating resources from 'cloudbuild.yaml': open /tmp/kustomization-2330757535/infra/cloudbuild.yaml: no such file or directory': must build at directory: not a valid directory: evalsymlink failure on '/tmp/kustomization-2330757535/infra/cloudbuild.yaml' : lstat /tmp/kustomization-2330757535/infra/cloudbuild.yaml: no such file or directory
   ```
   Changing the name of the file to `cloudbuildtrigger.yaml` fixed this error.
- Add annotation for keeping cloud resources even if their counterparts are deleted from the k8s cluster. This is done to prevent the deletion of cloud resources if the cluster goes down ([docs](https://cloud.google.com/config-connector/docs/how-to/managing-deleting-resources#keeping_resources_after_deletion)).


vedantthapa:
@simardeep1792 Feel free to add sops encrypted secrets to the `infrastructure/` directory once this is merged.

Stephen-ONeil (approved):
Approved, but just to confirm:

> Add annotation for keeping cloud resources even if their counterparts are deleted from the k8s cluster. This is done to prevent the deletion of cloud resources if the cluster goes down ([docs](https://cloud.google.com/config-connector/docs/how-to/managing-deleting-resources#keeping_resources_after_deletion)).

This also means that we will have to manually delete outgoing resources? If so, we'll need to document that and stay on top of it, or things could get messy.

vedantthapa:
@Stephen-ONeil 
Another way is to remove the annotation for a specific resource and then actually delete it from the cluster, which, will also clean it up on the cloud?

It's probably a bit tedious, but that's on purpose just so that there are guardrails around deleting any infrastructure.

In any case, I agree to your point on documenting this.

Stephen-ONeil:
> Alternatively, we can have this annotation on specific resources like cloud kms / buckets? All others follow the "gitops" lifecycle?

This sounds good to me!

vedantthapa:
Annotation for deletion policy is now only added to kms and bucket resources. See https://github.com/PHACDataHub/cpho-phase2/pull/221/commits/d70116eaa69b25745e19666967dfd25070fb37e7 for details.

simardeep1792 (commented):
LGTM!

PR #224, opened 2024-02-20
Metadata review and approval

msarar:
Added metadata, benchmarking and trend analysis approval logic


server/cpho/models/indicators.py
```diff
@@ -605,6 +635,26 @@ class Meta:
...
+            return SUBMISSION_STATUSES.NOT_YET_SUBMITTED
+
+        if self.last_version_id == self.last_submitted_version_id:
+            return SUBMISSION_STATUSES.SUBMITTED
+
+        if self.last_version_id == self.last_program_submitted_version_id:
+            return SUBMISSION_STATUSES.PROGRAM_SUBMITTED
+
+        return SUBMISSION_STATUSES.MODIFIED_SINCE_LAST_SUBMISSION
+
```
AlexCLeduc, line 657: This is repeated 3 times. Can this just be consolidated into a mixin, say `SubmissionHelpersMixin` ? 

I think we can keep the other repeated stuff, e.g. the `objects = ... active_objects = ` because managers vs. inheritance can get tricky, but this method seems like it's worth consolidating, and there may be others

AlexCLeduc (approved):
Looks great, thanks!

Issue #218, opened 2024-02-16
Use IaD for Cloud Armor Security Policy


vedantthapa:
Currently, regional cloud armor security policy resource is not supported in config connector so we resorted to using `gcloud` commands in `./k8s/Taskfile.yaml` to create / configure it.

This will be changed once the resource is available upstream.

PR #225, opened 2024-02-20
Configure cloud armor policy

vedantthapa:
- Add `ComputeBackendService` resource to bind cloud armor policy to the backend service at an infrastructure level.
   This resource was imported into KCC using the method described [here](https://cloud.google.com/config-connector/docs/how-to/import-export/export#exporting_example). This is a workaround as, [GKE Gateway API doesn't support configuring cloud armor when a regional XLB class is used](https://cloud.google.com/kubernetes-engine/docs/how-to/gatewayclass-capabilities#additional-services).
   In the future, if cloud armor is still unsupported, it'd be better to have this link up configured while provisioning a regional security policy resource. KCC doesn't have this resource yet (See https://github.com/PHACDataHub/cpho-phase2/issues/218).
- Add `GCPBackendPolicy` to enable HTTP logging for the backend service.
  Might be helpful to reduce the `sampleRate` at some point.
- Tune WAF rules for SSO.

Note that, the backend service is controlled by the GKE's gateway configuration (`xlb-` prefixed manifests in `./k8s/istio-ingress/`) and must be modified carefully to avoid a conflicting state since two manifests are reconciling this resource simultaneously -
- GKE gateway configuration
- `ComputeBackendService` IaD configuration

If the gateway configuration is modified, the IaD spec must be regenerated at `./infrastructure/backend-service.yaml`.

### DIFF
--- a/infrastructure/backend-service.yaml
+++ b/infrastructure/backend-service.yaml
@@ -1,0 +1,27 @@
+# Due to limitations of GKE's Gateway API, a cloud armor security policy for the
+# regional XLB's backend service cannot be configured via the k8s gateway spec.
+# The workaround, therefore, is to add the policy to the backend-service directly by
+# importing the backend service resource into KCC.
+
+# This file is not synced by Flux
+---
+apiVersion: compute.cnrm.cloud.google.com/v1beta1
+kind: ComputeBackendService
+metadata:
+  annotations:
+    cnrm.cloud.google.com/project-id: pht-01hp04dtnkf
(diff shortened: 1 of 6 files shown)


Stephen-ONeil (approved):
Thanks for figuring that out! Oof though, definitely something to bug Archy about, there's a _lot_ lacking in GCP when it comes to regional load balancer security (no regional managed certs, all these small gotchas with the XLB, etc).

PR #227, opened 2024-02-21
Fix sync for `infrastructure/`

vedantthapa:
Apparently when a dry run is executed by Flux's kustomize controller to apply resources, KCC _thinks_ that Flux is trying to modify an annotation (`cnrm.cloud.google.com/state-into-spec`) and denies the request because the value for that annotation is immutable. This consequently causes the reconciliation to fail.

This PR addresses the issue by explicitly adding the annotation to specific resources that KCC complains about. Some info on what this means and why it's not a disruptive change: https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/669#issuecomment-1339614273.

Essentially we're just setting a default value explicitly.

### DIFF
--- a/infrastructure/artifactregistry.yaml
+++ b/infrastructure/artifactregistry.yaml
@@ -3,6 +3,8 @@ kind: ArtifactRegistryRepository
 metadata:
   name: hopic-k8s-images
   namespace: cnrm-system
+  annotations:
+    cnrm.cloud.google.com/state-into-spec: merge
 spec:
   description: Artifact registry for HoPiC docker images
   format: DOCKER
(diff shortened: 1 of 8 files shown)


Stephen-ONeil (approved):
👍 

Was it easy to dig up the solution? If it's not a trivial google away, we should document it somewhere. Could be a comment in every file, or maybe start putting together a infrastructure dir README for this kind of stuff.

simardeep1792 (approved):

vedantthapa:
> Was it easy to dig up the solution? If it's not a trivial google away, we should document it somewhere. Could be a comment in every file, or maybe start putting together a infrastructure dir README for this kind of stuff.

Yep! Flux alerts were very vocal about it lol. 

I'll be updating the `./k8s/README.md` with gotchas and maintenance tasks soon.

PR #228, opened 2024-02-21
CNPG changes to log transactions

simardeep1792:
Enable Comprehensive SQL Logging in PostgreSQL Configuration

A change in the PostgreSQL configuration to enable logging of all SQL statements, including transactions. By setting `log_statement` to `'all'`, to enhance database's auditability and observability, ensuring capture comprehensive logs for all operations. This adjustment is crucial for troubleshooting and monitoring the database more effectively.

```yaml
log_statement: 'all' # Added to log all SQL statements, including transactions.
```

### DIFF
--- a/k8s/server/base/postgres/pg-cluster.yaml
+++ b/k8s/server/base/postgres/pg-cluster.yaml
@@ -30,6 +30,7 @@ spec:
       pgaudit.log_catalog: "off"
       pgaudit.log_parameter: "on"
       pgaudit.log_relation: "on"
+      log_statement: 'all' # Added to log all SQL statements, including transactions.
 
   nodeMaintenanceWindow:
     inProgress: false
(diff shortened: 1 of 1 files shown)


Stephen-ONeil (changes requested):
I went back to my original postgres logging PR, because I felt like there was a specific combination of `pgaudit.log` and `log_statement` that I worked out as being "best".

Looks like I went with [this](https://github.com/PHACDataHub/cpho-phase2/blob/1da60b13fb239651979aa00356a6df4cbeb46640/deploy/gcloud_init_setup.sh#L217):
```
"pgaudit.log=all,-misc"
...
# keep statement logging (DDL, queries) off, already captured better by pgaudit
"log_statement=none" 
```
Partially explained by the comment I left, but the idea was that a) pgaudit's logging is preferable to the standard logs, and b) we'd like to avoid overlapping logs. Note that `log-statement=none` doesn't totally disable logging, [depending on the `log_min_messages` value](https://www.postgresql.org/docs/current/runtime-config-logging.html#GUC-LOG-STATEMENT).

The combination in this PR has 
```
pgaudit.log: "ddl, role" # From "all, -misc" to specific events like DDL and role changes
...
log_statement: 'all' # Added to log all SQL statements, including transactions.
```
but that means a) we don't get the pgaudit version of transaction logs, and b) we get ddl and role logs from both pgaudit and the standard log messages.

That said, I've not looked at this stuff since back in August so if you have sources/reasons to say otherwise then let me know for sure!

Stephen-ONeil (approved):

PR #231, opened 2024-02-22
changed text colour to black to meet colour contrast requirements

kdompaul:
In order to meet the 4.5:1 colour contrast requirements I adjusted the text colour for the cyan and yellow backgrounds to black and kept the white text for the green background


msarar (commented):
Contents of this PR is already in https://github.com/PHACDataHub/cpho-phase2/pull/235
is this an error?

PR #247, opened 2024-03-01
feat: added table_title_overall and changed impact text to rich text

msarar:


server/cpho/models/indicators.py
```diff
@@ -172,8 +172,9 @@ class Indicator(models.Model, SubmissionHelpersMixin):
 
     # GENERAL
     measure_text = fields.TextField(null=True, blank=True)
-    impact_text = fields.TextField(null=True, blank=True)
     title_overall = fields.TextField(null=True, blank=True)
+    table_title_overall = fields.TextField(null=True, blank=True)
+    impact_text = RichTextField(config_name="notes", null=True, blank=True)
```
AlexCLeduc, line 177: I should have pointed this out earlier, but it would be good to make our own subclass of the RichTextField (like we do with all other fields in `fields.py`) so that we don't end up with silly migrations. It doesn't seem to matter much yet but we'll probably want to add labels via verbose_names

AlexCLeduc (approved):

PR #249, opened 2024-03-11
HSO only fields

msarar:
HSO only field tags and conditional rendering, 
Disabling indicator critical information from non privileged user

### DIFF
--- a/server/cpho/jinja2/indicators/_indicator_form.jinja2
+++ b/server/cpho/jinja2/indicators/_indicator_form.jinja2
@@ -1,7 +1,9 @@
 {{ form.media }}
 
 {% macro form_field(field, rich=false) %}
-  {% if (field.name in form.hso_only_field_names and respects_rule('is_admin_or_hso') or (field.name not in form.hso_only_field_names)) %}
+  {% set hso_field_display = field.name in form.hso_only_field_names and respects_rule('is_admin_or_hso') %}
+  {% set non_hso_field_display = field.name not in form.hso_only_field_names %}
+  {% if (hso_field_display or non_hso_field_display) %}
     <div class="row g-3 form-group mb-3">
       <div class="col-lg-3">
         <div>
--- a/server/cpho/jinja2/indicators/_indicator_form.jinja2
+++ b/server/cpho/jinja2/indicators/_indicator_form.jinja2
@@ -1,18 +1,26 @@
 {{ form.media }}
 
 {% macro form_field(field, rich=false) %}
-  <div class="row g-3 form-group mb-3">
-    <div class="col-lg-3">
-      <div class="row">{{ field.label_tag(attrs={'class': 'col col-form-label', 'aria-label': field.label }) }}</div>
+  {% if (field.name in form.hso_only_field_names and respects_rule('is_admin_or_hso') or (field.name not in form.hso_only_field_names)) %}
+    <div class="row g-3 form-group mb-3">
+      <div class="col-lg-3">
+        <div>
+          <div class="d-inline">{{ field.label_tag(attrs={'class': 'col col-form-label', 'aria-label': field.label }) }}</div>
+          {% if field.name in form.hso_only_field_names %}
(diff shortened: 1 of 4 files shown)


server/cpho/views/indicators.py
```diff
@@ -43,12 +43,34 @@ class IndicatorForm(ModelForm):
...
+            "detailed_indicator",
+            "sub_indicator_measurement",
+            "relevant_dimensions",
+            "relevant_period_types",
+        ]
+        if user and not test_rule("is_admin_or_hso", user):
+            for field in non_hso_readonly_fields:
+                # self.fields[field].disabled = True
+                # self.fields[field].widget.attrs["readonly"] = True
+                self.fields[field].widget.attrs["disabled"] = True
```
AlexCLeduc, line 69: use [field.disabled](https://docs.djangoproject.com/en/5.0/ref/forms/fields/#disabled) instead. Setting this on the widget will only set the attr on the frontend, but `field.disabled` will do this _and_ ignore incoming input values making it impossible to submit data anyway

Also, rm comments

server/cpho/jinja2/indicators/_indicator_form.jinja2
```diff
@@ -1,18 +1,26 @@
...
-    <div class="col-lg-3">
-      <div class="row">{{ field.label_tag(attrs={'class': 'col col-form-label', 'aria-label': field.label }) }}</div>
-    </div>
-    <div class="col-lg-9">
-      {% if rich %}
-        {{ field|safe }}
-      {% else %}
-        {{ field }}
-      {% endif %}
+  {% if (field.name in form.hso_only_field_names and respects_rule('is_admin_or_hso') or (field.name not in form.hso_only_field_names)) %}
```
AlexCLeduc, line 4: Took me a while to understand this condition. Can we use intermediary variables to clean it up. Can also be simplified because admins can see all fields
```jinja2
{% set is_hso_only = field.name in form.hso_only_field_names %}
{% set is_admin = respects_rule('is_admin_or_hso') %}
{% if is_admin or not is_hso_only %}
 ....
{% endif %}
```

server/cpho/views/indicators.py
```diff
@@ -76,11 +98,21 @@ class Meta:
...
+    )
+
     # make it display name_en attribute of DimensionType model
-    relevant_dimensions = RelevantDimensionChoiceMultiCheckbox(
+    relevant_dimensions = ModelMultipleChoiceFieldWithTranslation(
         required=False,
         queryset=DimensionType.objects.all(),
-        widget=forms.CheckboxSelectMultiple(),
+        widget=forms.CheckboxSelectMultiple(
+            # attrs={"class": "form-check-input"}
```
AlexCLeduc, line 114: more comments

server/cpho/views/indicators.py
```diff
@@ -32,7 +32,7 @@
 
 
 # might need to move this to a form_fields.py file
-class RelevantDimensionChoiceMultiCheckbox(forms.ModelMultipleChoiceField):
+class ModelMultipleChoiceFieldWithTranslation(forms.ModelMultipleChoiceField):
```
AlexCLeduc, line 35: 👍

server/cpho/views/indicators.py
```diff
@@ -43,12 +43,34 @@ class IndicatorForm(ModelForm):
...
-        }
+        # widgets = {
+        #     "relevant_period_types": forms.CheckboxSelectMultiple,
+        # }
+
+    def __init__(self, *args, **kwargs):
+        user = None
+        if "user" in kwargs:
+            user = kwargs.pop("user", None)
+        super(IndicatorForm, self).__init__(*args, **kwargs)
```
AlexCLeduc, line 54: super with arguments is an outdated python2 thing and an anti-pattern in python3 (unless you want to wacky things like skip the parent class to call the grandparent). Just use `super().__init__(*args,**kwargs)`

AlexCLeduc (approved):

PR #250, opened 2024-03-11
updates global status even if no data for bench or trend

msarar:
updates global status for metadata status even if no data for bench or trend.

This is because sometimes indicators have no benchmarking or trend data but would still need to be approved


AlexCLeduc (approved):
👍

Issue #133, opened 2023-10-06
Load test the k8s configuration


Stephen-ONeil:
Doesn't need to be very fancy or _too_ realistic, although if you pick a solution that can manage some automation then I'd recommend talking to @AlexCLeduc and hashing out some test credentials and basic workflows to properly put the app nodes through their paces.

Ideally pick a tool where we can re-run the load balancing whenever we want, as we'll be needing load testing at least until the configuration is fully stable (and even then it could be useful to the devs if the app requirements include heavier routes down the line, etc).

We want to do this soon to put the experimental-ish [1-1 app node to django process](https://github.com/PHACDataHub/cpho-phase2/pull/129) configuration through its paces. It works when it's just me poking around the test deployment, but it's definitely suboptimal for higher traffic volume. Adding some sort of request counting to the horizontal scaling and making sure the load balancing is smart could be necessary, but we should see how bad it is to start with before going down the rabbit hole. Compare it to nodes with more resources each running multiple django process via gunicorn.

PR #251, opened 2024-03-12
Pre-prod performance tuning

Stephen-ONeil:
Leaving our server pods on the minimal resources for now, but increasing the gunicorn workers per-pod (from 1 to 3). I expect this to fit based on pre-prod resource usage metrics and local testing, but it'll be something to keep an eye on. Due to the long running data upload requests, concurrency and scaling is necessary to keep the site from going unresponsive under regular use.

We'll also need to tune the cluster appropriately down the road (or maybe the per-pod resources), depending on if this puts the average CPU above the scaling threshold. Expect this to be fine though. @vedantthapa tagging you FYI.

Related #133


AlexCLeduc (approved):
Cool!

PR #252, opened 2024-03-12
Infobase export

AlexCLeduc:
Uses phac-helpers' excel utilities to easily produce a 4-sheet (inds, ind-data, benchmarking, trends) semi-normalized export of all relevant data. 

Here's how you access it:

Screenshot 2024-03-12 at 16 23 06

### DIFF
--- a/server/cpho/views/infobase_export.py
+++ b/server/cpho/views/infobase_export.py
@@ -1,0 +1,96 @@
+from django.core.exceptions import PermissionDenied
+from django.http import HttpResponse
+from django.utils.functional import cached_property
+from django.views import View
+
+import openpyxl
+from phac_aspc.django.excel import (
+    AbstractExportView,
+    AbstractSheetWriter,
+    ChoiceColumn,
+    Column,
+    CustomColumn,
+    ModelColumn,
+    ModelToSheetWriter,
+)
+from phac_aspc.rules import test_rule
+
+from cpho.models import Benchmarking, Indicator, IndicatorDatum, TrendAnalysis
+
+
+# indicator_columns = []
+class IndicatorSheetWriter(AbstractSheetWriter):
+    sheet_name = "indicator"
+
+    def get_column_configs(self):
+        model = self.iterator.model
+
+        fields_to_write = model._meta.fields
+        return [
+            ModelColumn(model, field.name)
+            for field in fields_to_write
+            if field.name != "relevant_period_types"
+        ]
+
+
+class IndicatorDatumSheetWriter(ModelToSheetWriter):
+    sheet_name = "indicator_data"
+
+
+class BenchmarkingSheetWriter(ModelToSheetWriter):
+    pass
+
+
+class TrendSheetWriter(ModelToSheetWriter):
+    pass
+
+
+class InfobaseExportView(View):
+
+    @cached_property
+    def workbook(self):
+        return openpyxl.Workbook(write_only=True)
+
+    def write_indicators(self):
+        # todo: use submitted indicators instead
+        writer = IndicatorSheetWriter(
+            workbook=self.workbook, iterator=Indicator.objects.all()
+        )
+        writer.write()
+
+    def write_indicator_data(self):
+        writer = IndicatorDatumSheetWriter(
+            workbook=self.workbook, iterator=IndicatorDatum.objects.all()
+        )
+        writer.write()
+
+    def write_trends(self):
+        writer = TrendSheetWriter(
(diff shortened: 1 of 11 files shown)


server/cpho/views/infobase_export.py
```diff
@@ -0,0 +1,220 @@
...
+    ChoiceColumn(IndicatorDatum, "data_quality"),
+    ChoiceColumn(IndicatorDatum, "reason_for_null"),
+    ModelColumn(IndicatorDatum, "value"),
+    ModelColumn(IndicatorDatum, "value_lower_bound"),
+    ModelColumn(IndicatorDatum, "value_upper_bound"),
+    ChoiceColumn(IndicatorDatum, "value_unit"),
+    ChoiceColumn(IndicatorDatum, "value_displayed"),
+    ModelColumn(IndicatorDatum, "single_year_timeframe"),
+    ModelColumn(IndicatorDatum, "multi_year_timeframe"),
+]
```
AlexCLeduc, line 124: Here's how columns are defined

server/cpho/views/infobase_export.py
```diff
@@ -0,0 +1,220 @@
...
+    @cached_property
+    def workbook(self):
+        return openpyxl.Workbook(write_only=True)
+
+    def write_indicators(self):
+        # todo: use submitted indicators instead
+        writer = IndicatorSheetWriter(
+            workbook=self.workbook, iterator=Indicator.objects.all()
+        )
+        writer.write()
```
AlexCLeduc, line 183: TODO later: target submitted versions, or perhaps make this a query-param so infobase can test non-submitted data

server/cpho/views/infobase_export.py
```diff
@@ -0,0 +1,220 @@
...
+
+    def get_column_configs(self):
+        return indicator_columns
+
+
+indicator_data_columns = [
+    ModelColumn(IndicatorDatum, "id"),
+    CustomColumn("indicator_id", lambda x: x.indicator_id),
+    # indicator columns useful to help identify??
+    CustomColumn("indicator_id", lambda x: x.indicator.name),
```
msarar, line 87: should these be "indicator_name"?

server/cpho/views/infobase_export.py
```diff
@@ -0,0 +1,220 @@
...
+class IndicatorSheetWriter(AbstractSheetWriter):
+    sheet_name = "indicator"
+
+    def get_column_configs(self):
+        return indicator_columns
+
+
+indicator_data_columns = [
+    ModelColumn(IndicatorDatum, "id"),
+    CustomColumn("indicator_id", lambda x: x.indicator_id),
```
msarar, line 85: CustomColumn("indicator_id"

search returns 6 instances of these, i think these are copy/paste errors.
We need to give them the right column names

server/cpho/views/infobase_export.py
```diff
@@ -0,0 +1,220 @@
...
+    ChoiceColumn(Benchmarking, "labels"),
+]
+
+
+class BenchmarkingSheetWriter(ModelToSheetWriter):
+    def get_column_configs(self):
+        return benchmarking_columns
+
+
+trend_columns = [
```
msarar, line 157: We might need to add indicator metadata columns here especially the indicator_id, 

We could also add the following (indicator.name,  indicator.detailed_indicator, x.indicator.sub_indicator_measurement) for better readability of the file

msarar (approved):
Thanns Alex
This looks really good.


This gives me an idea for a 'nice to have' feature for the future:
Users could select the indicators they want (using a checkbox or sth) of which they can export all data for the selected indicators. We could use most of the code here for that. However the only thing is that in that case the export will be .xslx which they wont be able to upload to the system as the upload only supports .csv. But it'd still be a nice feature to have if we want/ see a need for it.

AlexCLeduc:
> Thanns Alex This looks really good.
> 
> This gives me an idea for a 'nice to have' feature for the future: Users could select the indicators they want (using a checkbox or sth) of which they can export all data for the selected indicators. We could use most of the code here for that. However the only thing is that in that case the export will be .xslx which they wont be able to upload to the system as the upload only supports .csv. But it'd still be a nice feature to have if we want/ see a need for it.

Yeah I was hoping we'd use the phac-helpers for the CSV export but the tool wasn't very mature at the time, and didn't yet support CSV.  The phac-helpers now support CSV and excel, so we can re-use the same column definitions, but I'd rather not break anything at the moment. 

Also note that we're going to tailor this to infobase's specs, which might be different than the average user. Ideally we could re-use column definitions, but I think we'd have different exports for average users.

Issue #64, opened 2023-07-14
Production DB management mechanism & playbook 


Stephen-ONeil:
We need to keep the production database fenced off and assure its security and integrity, but we'll also need a mechanism/playbook for executing data management scripts and (non-trivial) migrations against it. Acceptance criteria will mean something that can **pass a protected-B level security assessment** while also being **sufficiently flexible and confidence-inspiring for devs**.

The GitOps/zero trust ideal would be to carefully plan out django migrations, and to use data migrations instead of scripts, so that they can be run hands-off in the cloud as part of the project's CD infrastructure. This no-one-touches-the-database approach theoretically breezes past a prot-B assessment (as much as anything can breeze past the gauntlet).

Fully hands-off DB management is less flexible and seems to not inspire developer confidence. To my understanding, lack of confidence in the approach is in large part because no hands-on-keyboard means no ability to steer the ship when things go sideways.

Personally, I'd like to try and work out the hands-free approach, but I also think some sort of break-glass direct access system for emergencies is sensible. Of course, that means doing all the work and maintenance for both (although it'd be easier to get break-glass console access past security then saying it will be the default management method).

Stephen-ONeil:
@AlexCLeduc I'm going to defer to you, as the Chief Django Officer. We've had this debate enough times, haha.

Stephen-ONeil:
Just keep in mind that, if not on this project then on something needing protected-B, you might need to "secure" the manual developer access based approach by smothering it in procedure and bureaucracy.

Stephen-ONeil:
Misc links on migrations:
- [Database migrations in cloud run jobs](https://cloud.google.com/blog/topics/developers-practitioners/running-database-migrations-cloud-run-jobs)
- [Django on Cloud Run (cloudmigrate.yaml)](https://cloud.google.com/python/django/run#:~:text=The%20cloudmigrate.yaml%20file%20performs)

Stephen-ONeil:
Should note, with the current configuration, Cloud SQL Auth Proxy can't connect from dev machines, because the DB can only be reached from inside the VPC. Don't want to peer our dev machines to the VPC. A short-term solution for using Cloud SQL Auth Proxy could be to write a connection script that includes adding a temporary add public IP to the DB, allow listed to the IP of the dev running the script.

[...]

PR #256, opened 2024-03-14
Server management session script

Stephen-ONeil:
Part 1 of #64

Leverages `kubectl debug`, which lets you attach an ephemeral container to an existing container's process pool. This script just puts connecting to the cluster and selecting from the available server pods on-rails (also clones the server process' env vars to the attached debug/management shell, which is a minor trick but would be annoying to do manually every time).

As a follow up to this, we can strip down the prod image to the absolute minimum (possibly even "distroless") and add a dedicated management/debug image to the build pipeline. The script assumes we'll do that and push both containers with different names and the same per-commit/branch tags, to correlate the versions.

You can copy files to/from the running maintenance containers with `kubtectl cp`, similar in concept to `scp` in an SSH based world.

@AlexCLeduc could you pull this down and try the script? Should be POSIX and work on Mac, but worth double checking. Window users may or may not need to run it from a GCP cloud shell, but it should work fine there too. 

@vedantthapa tagging FYI

### DIFF
--- a/k8s/server_management_session.sh
+++ b/k8s/server_management_session.sh
@@ -1,0 +1,85 @@
+#!/usr/bin/env bash
+set -o errexit
+set -o pipefail
+set -o nounset
+
+PROJECT_ID=pht-01hp04dtnkf
+REGION=northamerica-northeast1
+CLUSTER=hopic-cluster
+SERVER_CONTAINER_NAME=server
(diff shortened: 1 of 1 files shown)


AlexCLeduc (approved):
Thanks for testing on cloudshell

PR #257, opened 2024-03-19
Distroless runtime and management images

Stephen-ONeil:
Switching the prod (and CI test runner) image to an ultra stripped down distroless image, adding a separate dev/management image for use with the script in #256.

I've done very minimal hands-on testing of the deployed distroless app image, but it seems to be fine and the tests all pass on the distroless test runner image in CI. If you want to test the distroless app instance, it has an ephemeral build reachable [here](https://distroless.dev.hopic-sdpac.phac-aspc.alpha.canada.ca/). You can use the management script on this branch to attach and add test user credentials to that ephemeral deployment :+1: 

Pre-merging check list, on this branch:
  - [x] Revert the `GITHUB_MAIN_BRANCH_NAME` value to `prod`
  - [x] Revert the ephemeral deployment image policy to `flux-system:server`
  - [x] Remove the redundant `securityContext` block in the ephemeral deployment
  
Post-merge check list:
  - [ ]  revert [this](https://github.com/PHACDataHub/cpho-phase2/commit/065ce36c80ac8021f336536446c075217f2ca3d4) commit on `prod` to clean up the ephemeral deployment


Stephen-ONeil:
Fixed the coverage file write access issue. The next mystery is why the coverage reported on this branch is roughly half of masters...

k8s/server/overlays/ephemeral/django/deployment.yaml
```diff
@@ -7,7 +7,10 @@ spec:
     spec:
       containers:
       - name: server
-        image: northamerica-northeast1-docker.pkg.dev/pht-01hp04dtnkf/hopic-k8s-images/cpho-phase2:prod-207fc8eb-1711642657 # {"$imagepolicy": "flux-system:server"}
+        image: northamerica-northeast1-docker.pkg.dev/pht-01hp04dtnkf/hopic-k8s-images/cpho-phase2:distroless-runtime-and-management-image-aec37750-1712599872 # {"$imagepolicy": "flux-system:distroless-server"}
+        securityContext:
```
vedantthapa, line 11: I noticed that the base deployment has a `securityContext` block, do we need to explicitly define it here as well?

vedantthapa (approved):
Other than a minor question, everything else looks good :)

Thanks!

AlexCLeduc (approved):
Tried the debug script and successfully got in there. You weren't kidding about this image being slim, I couldn't exec `sh` or even run `ls` in there!

PR #259, opened 2024-03-20
fix: formset error handling on non unique data entry

msarar:


AlexCLeduc (approved):
you'll probably need to update text, those text keys should exist already

PR #262, opened 2024-03-25
fix: trend analysis feedback fixes

msarar:
Added choice: Caution for trend data quality
Merged year and year range to one field and fixed validation rules
Trend segment validation update:  allows input like: "YYYY-YYYY to YYYY-YYYY" or "YYYY-YYYY"
added unit field


AlexCLeduc (approved):
Looks good, thanks

super nitpick: `map` is kind of weird, list comprehensions are more pythonic, up to you if you want to update

PR #267, opened 2024-04-04
moved hidden_a11y_label to macros file

andGuo:
- moved hidden_a11y_label to macros file
- removed accessible_label macro


AlexCLeduc (approved):
Also works well, thanks

PR #276, opened 2024-04-11
5767 rs restore benchmarking

msarar:
Scripts to mass populate benchmarking and trend data from respective csv files.
Also some form validation changes as working with the data provided more insight

Diabetes indicator might need updating: waiting on feedback from HSO


server/cpho/views/indicator_metadata.py
```diff
@@ -157,7 +157,14 @@ def clean_year(self):
         year = self.cleaned_data["year"]
 
         if year is None or year == "":
-            return None
+            if self.cleaned_data["oecd_country"].name_en in ["OECD"]:
+                return None
```
AlexCLeduc, line 161: Why don't OECD rows have years?

server/cpho/scripts/dev.py
```diff
@@ -26,7 +26,7 @@
 @transaction.atomic
 def run():
     create_admins()
-    create_data()
+    # create_data()
```
AlexCLeduc, line 29: is this supposed to go into main like this?

server/cpho/views/indicator_metadata.py
msarar: I asked HSO about this, they said some OECD rows don't have years
image

server/cpho/scripts/dev.py
msarar: good catch!
I had it commented out as I wanted to seed my local db with data from prod.
its better to revert this (even though we're not going to seed prod anyway)

AlexCLeduc (approved):
Thanks for accommodating this weird situation! It looks great at a glance, not gonna nitpick one-time scripts, let's just test it again locally before running anything in prod.

PR #281, opened 2024-04-17
Fix healthcheck 301s

vedantthapa:
- Increase `initialDelaySeconds` to `135` to delay readiness checks so that the `server` pod has more time to startup
- Fix readiness probe request headers for ephemeral deployments
- Exempt the `healthcheck/` endpoint from secure ssl redirects to fix 301s in the `server` pod


Stephen-ONeil (approved):
:+1: thanks!

PR #283, opened 2024-04-18
fix: disable legacy login

AlexCLeduc:
Looks like you did the same [here](https://github.com/PHACDataHub/cpho-phase2/compare/main...disable-login) @vedantthapa , we want this is in prod too. If you'd like to keep legacy login around to test the new domain, that's fine, as long as we turn this off when we finally launch.


vedantthapa (approved):
We've already tested the new domain, so all good on that end! 

Thanks :)

Issue #182, opened 2024-01-10
GKE upgrade breaks postgres cluster


vedantthapa:
There was an occurrence previously when a maintenance window was added to the GKE Autopilot Cluster. After maintenance was completed, the cloudnative-pg cluster ended up in a broken state. This might be due to the fact that cloudnative-pg operator expects to have `nodeMaintenanceWindow.inProgress` field set to `true` on the `Cluster` resource whenever there's a kubernetes upgrade (See [docs](https://cloudnative-pg.io/documentation/1.15/kubernetes_upgrade/#kubernetes-upgrade)).

We've removed the maintenance window for now so that no automated updates are triggered, however, a solution is needed to either automatically update the field in the `Cluster` resource before a GKE maintenance window and toggle it back to the orginial value once the upgrade is finished _OR_ alert the user to the availability of an upgrade so they can modify the field, start the upgrade, and then toggle the field back to its initial value.

vedantthapa:
Setting `nodeMaintenanceWindow.inProgress` to `true` might cause issues with the Cluster as per the warning mentioned on this page: https://cloudnative-pg.io/documentation/1.15/kubernetes_upgrade/#kubernetes-upgrade

vedantthapa:
Another idea is to set the maintenance window and have a github action workflow run on a schedule before and after the maintenance window to update the relevant field in the manifest.

PR #284, opened 2024-04-19
Add maintenance window for GKE

vedantthapa:
Apparently GKE doesn't upgrade the nodes immediately once the control plane has been upgraded and internally it depends on a lot factors. Check [this](https://www.googlecloudcommunity.com/gc/Google-Kubernetes-Engine-GKE/Autopilot-node-now-stuck-in-Ready-SchedulingDisabled-state/m-p/515896) out, it took almost 3.5 days for the nodes to finish upgrading after the control plane was upgraded.

This PR sets a maintenance window for the cluster during work hours -> 10AM - 4PM ADT every week on Wed, Thu and Fri. Date and time info in the `gcloud` command is in UTC. The changes have already been applied.

See https://cloud.google.com/kubernetes-engine/docs/concepts/maintenance-windows-and-exclusions for details.

Relates to #182

### DIFF
--- a/k8s/Taskfile.yaml
+++ b/k8s/Taskfile.yaml
@@ -11,6 +11,9 @@ tasks:
 
   infra-up:
     desc: Create infrastructure components upto the cluster-level
+    vars:
+      CURRENT_DATE:
+        sh: echo {{now | date "2006-01-02"}}
     cmds:
       # Enable APIs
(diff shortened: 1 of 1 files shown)


Stephen-ONeil (approved):
:+1: good catch!

PR #286, opened 2024-04-22
Independent approvals

msarar:
Updated get_submission_statuses that calculates hso and program submissions separately


server/tests/test_submission_statuses.py
```diff
@@ -84,102 +87,151 @@ def create_data(abc, xyz=NO_DATA_TRIPLE, ijk=NO_DATA_TRIPLE):
...
+            ijk_dim.id: SUBMISSION_STATUSES.MODIFIED_SINCE_LAST_SUBMISSION,
+        },
+        "program_global_status": SUBMISSION_STATUSES.MODIFIED_SINCE_LAST_SUBMISSION,
     }
 
 
-def test_modified_takes_precedence_over_program_submitted():
+def test_hso_submitted_with_edit():
     ind, period, abc_dim, xyz_dim, ijk_dim = create_data(
-        abc=[SUBMISSION_STATUSES.SUBMITTED] * 3,
```
AlexCLeduc, line 148: Is there a reason you replaced the constants with string literals? Is it that the options have changed?

AlexCLeduc (approved):
Works on my machine with admins and program users, thanks Raad!

server/tests/test_submission_statuses.py
msarar: Yes, 
Its because options have changed and program and hso statuses are independent now

For example,
SUBMISSION_STATUSES.PROGRAM_SUBMITTED used to mean that the data has been submitted by program but not by HSO
but now it means data has been submitted by program but it gives no insight into hso submission status.

PR #287, opened 2024-04-24
5769 kd benchmarking user view

kdompaul:
Worked on Users should be able to view benchmarking section but not be able to add or edit this data. HSO and Admin edit only.
added test_benchmarking_permissions test to meet threshold


msarar (approved):

server/cpho/rules.py
```diff
@@ -95,7 +95,14 @@ def can_export_benchmarking(user, indicator):
...
 def can_edit_benchmarking(user, indicator):
-    return can_edit_indicator(user, indicator)
+    return is_admin_or_hso(user)
+
+
+@auto_rule
+def can_view_benchmarking(user, indicator):
+    return can_edit_benchmarking(user, indicator) or (
+        is_inputting_user(user) and not is_admin_or_hso(user)
+    )
```
AlexCLeduc, line 105: This would allow other indicator groups to view benchmarking for indicators they don't usually have access to. I think we just want `return can_access_indicator(user,indicator)`

PR #292, opened 2024-04-25
feat: Streamlined periods

AlexCLeduc:
So I thought about adding a `settings.CURRENT_QUARTER`, but that would mean pushing an update each quarter (this will probably happen anyway, but it's annoying) and it means we can't have 2 quarters open at once, which may happen at some point as some programs are going to be more punctual than others. 

So instead, we have now a `is_current` bool column in the period table. On an indicator's "home" page, program users will only see the current periods (this should just be one, if `relevant_period_types` is properly set) in the table. 

For an indicator with  `relevant_period_types=['calendar_year']`, 
This is the HSO view:
Screenshot 2024-04-25 at 15 04 04

This is the program view:
Screenshot 2024-04-25 at 15 04 49


server/cpho/admin.py
```diff
@@ -6,4 +6,8 @@
 
 for model in this_app.values():
     if hasattr(model, "__add_to_admin"):
-        admin.site.register(model)
+
+        if hasattr(model, "adminClass"):
+            admin.site.register(model, model.adminClass)
+        else:
+            admin.site.register(model)
```
AlexCLeduc, line 13: allow custom admin classes

server/cpho/models/lookups.py
```diff
@@ -118,9 +120,10 @@ def __str__(self):
 
     @staticmethod
     def get_currently_relevant_periods():
-        return Period.objects.filter(
-            year__in=[settings.CURRENT_YEAR - 1, settings.CURRENT_YEAR]
-        )
+        return Period.objects.filter(is_current=True)
+
+    class adminClass(admin.ModelAdmin):
+        list_display = ["year", "year_type", "quarter", "is_current"]
```
AlexCLeduc, line 128: makes it easy to see which periods are active in the admin view

server/cpho/views/indicators.py
```diff
@@ -401,6 +403,7 @@ def get_context_data(self, **kwargs):
             "metadata_submission_statuses": get_metadata_submission_statuses(
                 indicator
             ),
+            "alternate_periods": alternate_periods,
```
AlexCLeduc, line 406: The idea here was to provide links for HSO to add data to adjacent years/quarters. Seems better to have this and not need it. There's no easy alternative way for HSO to add data in other periods

server/cpho/fixtures/periods.yaml
```diff
@@ -13,6 +13,7 @@
   fields:
     year: 2023
     year_type: calendar
+    is_current: True
```
AlexCLeduc, line 16: the fixtures were updated for dev purposes but I think I'll use the django-admin in prod

msarar (approved):
Looks Good!
Thanks Alex!

So the idea is we would have multiple is_active periods 
like 2023_Calendar. 2023_fiscal, 2023_fiscal_q4 and depending on the relevant periods of an indicator, show the the active relevant periods to the programs

the only thing i can think of is
lets say one indicator needs data for 2023 fiscal to be entered, but for another data for 2022 fiscal needs to be entered by the program.

setting them both as is_active means for all indicators programs can enter data for both periods (given the relevant periods is set to fiscal)

AlexCLeduc:
Solving that problem would involve setting current-ness at the indicator-period level, that sounds annoying. I am hoping if a period is still technically open but marked as approved, the program users won't be confused about which period needs data entry.

PR #298, opened 2024-04-29
fix: make health-check fail when there are pending migrations

AlexCLeduc:
The health-check will fail if there are un-applied migrations.

It also passes if migrations are ahead of the source code (e.g. when a new migration is deployed and applied, the old containers still pass this check)


AlexCLeduc:
Actually, this is pointless, the pods run migrations when they start so the HC will always pass

PR #302, opened 2024-05-01
fix: Benchmarking year form validation

msarar:
Benchmarking Year now expects MM/YYYY for Covid 19 Deaths


server/cpho/views/indicator_metadata.py
```diff
@@ -166,6 +163,32 @@ def clean_year(self):
                 )
                 return year
 
+        if self.cleaned_data["indicator"].id == 168:
+            year = str(year).strip()
+            pattern = r"^(\d{1,2})\s*\/\s*(\d{4})$"
+
+            match = re.match(pattern, year)
```
AlexCLeduc, line 170: can we extract this regex check to its own little function and unit-test it with a few cases?

AlexCLeduc (approved):

PR #304, opened 2024-05-06
fix: Read only form for indicator datum

msarar:
Indicator data becomes read only if:
user is program user and period is not current

### DIFF
--- a/server/cpho/fixtures/periods.yaml
+++ b/server/cpho/fixtures/periods.yaml
@@ -29,7 +29,7 @@
   fields:
     year: 2023
     year_type: fiscal
-    is_current: True
+    is_current: False
 - model: cpho.period
   pk: 7
   fields:
@@ -84,7 +84,7 @@
     year: 2023
     quarter: 1
     year_type: fiscal
-    is_current: True
+    is_current: False
 - model: cpho.period
   pk: 16
   fields:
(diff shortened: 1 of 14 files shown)


server/cpho/fixtures/periods.yaml
```diff
@@ -103,3 +103,18 @@
...
+    year_type: calendar
+    is_current: True
+
+- model: cpho.period
+  pk: 18
+  fields:
+    year: 2024
+    quarter: 1
+    year_type: fiscal
+    is_current: True
```
AlexCLeduc, line 120: CY 2024 should be pk:19 and FY 2024-Q1 should be 20

server/cpho/jinja2/benchmarking/manage_benchmarking_data.jinja2
```diff
@@ -84,7 +84,8 @@
 
   <script>
     let addButton = document.querySelector("#add-form");
-    addButton.addEventListener('click', addForm);
+    if (addButton)
+      addButton.addEventListener('click', addForm);
```
AlexCLeduc, line 88: nitpick: add curly brackets for if statements in javascript

server/tests/test_manual_data_input.py
```diff
@@ -13,11 +13,15 @@ def test_predefined_create_from_scratch(vanilla_user_client):
     url = reverse(
         "manage_indicator_data", args=[ind.id, period.id, sex_cat.pk]
     )
-    with patch_rules(can_edit_indicator_data=True):
```
AlexCLeduc, line 16: Idk if we're testing this elsewhere, but we should test the case in which can_edit is False and can_view is true, so that the read-only case is covered.

AlexCLeduc (approved):

PR #309, opened 2024-05-07
chore: prod dumps for indicator directory and indicator metadata

msarar:
Loaddata was successfully able to ingest these fixtures


AlexCLeduc (approved):

AlexCLeduc:
I'm curious why you timestamped and added new files? I would prefer to just replace the files rather than accumulating timestamped versions, git will keep track of the dates and we can keep using the same loaddata commands in the README

msarar:
> I'm curious why you timestamped and added new files? I would prefer to just replace the files rather than accumulating timestamped versions, git will keep track of the dates and we can keep using the same loaddata commands in the README

We are actually not using the fixtures for indicator metadata and directories in the README
but I agree we dont need to keep timestamped versions of both files as git already does it for us.
fixing and pushing soon!

PR #311, opened 2024-05-09
chore: deprecate alpha dns zone

vedantthapa:
Since we've transitioned to the `prod` domain (https://github.com/PHACDataHub/cpho-phase2/pull/285) and have been stable for a while it'd be good to clean up the alpha DNS Managed Zone resource and it's record sets.

### DIFF
--- a/infrastructure/dns-zone.yaml
+++ b/infrastructure/dns-zone.yaml
@@ -42,24 +42,6 @@ spec:
   location: northamerica-northeast1
   networkTier: STANDARD
 ---
-# ephemeral DNS config
-apiVersion: dns.cnrm.cloud.google.com/v1beta1
-kind: DNSRecordSet
-metadata:
-  name: hopic-ephemeral-dns-record-set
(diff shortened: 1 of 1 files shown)


vedantthapa:
Hmm, good point! 

We can have another directory in `./k8s/server/overlays` called `dev` with the alpha domain SSO secrets to host a build, however, I'm not sure if we want to have a 24/7 running dev environment. Although doesn't have to be always running, could treat it with an ephemeral nature with the benefit -  _this one has a functioning SSO_?

Stephen-ONeil:
> Hmm, good point!
> 
> We can have another directory in `./k8s/server/overlays` called `dev` with the alpha domain SSO secrets to host a build, however, I'm not sure if we want to have a 24/7 running dev environment. Although doesn't have to be always running, could treat it with an ephemeral nature with the benefit - _this one has a functioning SSO_?

Sounds good to me!

AlexCLeduc (approved):

PR #316, opened 2024-05-14
fix: added age group custom sorting

msarar:


server/cpho/jinja2/indicator_data/manage_indicator_data.jinja2
```diff
@@ -200,7 +200,14 @@
           </thead>
           <tbody id="{{ dimension_type }}-table-body"
                  class="text-center age-group-form-list">
-            {% for form in age_group_formset %}{{ age_group_form(form) }}{% endfor %}
+            {% for item in age_group_sorted %}
+              {% for form_instance in age_group_formset %}
+                {% if form_instance.instance.id == item.id %}
+                  {% set form = form_instance %}
+                  {{ age_group_form(form) }}
```
AlexCLeduc, line 207: suuuper nitpick, is it worth creating a variable? Why not just `age_group_form(form_instance)`

AlexCLeduc: Have you tested this with new records that are invalid? E.g. POST a form with new invalid records. They wouldn't be in the querset, are they still rendered?

AlexCLeduc: Yeah this is a bug, maybe we can follow the loop with this one,
```jinja2
{% for form_instance in age_group_formset %}
  {% if not form_instance.instance.id %}
    {{age_group_form(form_instance)}}
  {% endif %}
{% endfor %}
```

msarar: Good catch
i pushed up the changes
Thanks
:)

server/cpho/views/view_util.py
```diff
@@ -169,6 +171,33 @@ def check_rule(self):
...
+                ).zfill(6)
+            else:
+                return str(int(first_numeric.replace("+", "").strip())).zfill(
+                    6
+                )
+        return ind_data.literal_dimension_val
+
+    else:
+        raise ValueError("Custom Sort: Dimension type is not age")
+
```
AlexCLeduc, line 195: This is still pretty confusing, can we 

1. Rename this so make it clear this outputs a "sort" score, e.g. `def age_group_sortable_score(indicator_datum)` for a single indicator_datum
2. Remove the dimension_type check, which could cause N+1 problems (we shouldn't need to fetch dimension types on all indicator-datum records, that seems like overkill). The caller should be responsible in making sure they're not passing non-age-group data
3. Write some tests around this function or at least the sort function. Ideally we'd have a bunch of 2-element cases.

AlexCLeduc (approved):
Thanks, looks great!

PR #321, opened 2024-05-17
fix: Arrow flags

msarar:


AlexCLeduc (approved):
LGTM

Sorry about the wait, this went unnoticed after the long weekend. Please bug me if I take more than a day or two 😄

PR #327, opened 2024-05-27
feat: add year-filterable datum changelog

AlexCLeduc:
This is just the global changelog but just for the indicator-datum model, and you can add `?year=2023` on the URL to fitler to a particular year


AlexCLeduc:
2nd commit is somewhat unrelated, HSO just asked for it. We forgot to add the extra models to the single-indicator changelog. TrendAnalysis has also always been missing from the changelogs, so that's been added in.

msarar (approved):

PR #328, opened 2024-05-27
fix: revamped validation for upload

msarar:
Added check for closed periods.
Also,
Added a class "IndDataCleanMixin"
that helps with validation for indicator datum.
Methods in this class expects input for field and returns validation errors if appropriate.

Reason for making this a class of its own: So that both IndicatorDatumForm and Upload can use the same class to centralize validation logic.

Also reformatted some of the upload code to make it more readable

### DIFF
--- a/server/cpho/views/view_util.py
+++ b/server/cpho/views/view_util.py
@@ -120,6 +120,57 @@ class BaseInlineFormSetWithUniqueTogetherCheck(BaseInlineFormSet):
                         raise ValidationError(error_msg)
 
 
+class IndDataCleanMixin:
+    """
+    Mixin for cleaning indicator data values
+    Shared data validation logic between the indicator data form and the indicator data upload.
+    """
+
+    def IndDataCleanValue(self, value, value_unit):
+        if value:
+            value = float(value)
+            if value < 0:
+                return tm("value_cannot_be_negative")
+            if value_unit == "percentage":
+                if value and (value > 100 or value < 0):
+                    return tm("percentage_out_of_bounds_err")
+        return None
+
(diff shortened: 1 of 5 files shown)


server/cpho/views/view_util.py
```diff
@@ -120,6 +120,57 @@ def clean(
...
+                return tm("value_cannot_be_negative")
+            if value_unit == "percentage":
+                if value and (value > 100 or value < 0):
+                    return tm("percentage_out_of_bounds_err")
+        return None
+
+    def IndDataCleanValueLower(self, value, value_lower):
+        if (value and value_lower) and float(value) < float(value_lower):
+            return tm("lower_bound_must_be_lower_than_value")
+        return None
```
AlexCLeduc, line 142: These are a bunch of independant functions, this class is more of a namespace than something you mixin or instantiate. Classes as namespaces are an OK pattern, usually they look something like this

```python
class MyHelperFunctions:

    @staticmethod
    def process_thing(x):
       # helpers without dependencies on other helpers are staticmethods
       return ...

    @staticmethod
    def _common_process_thing
        # private methods can be used by the other helpers but are discouraged by end-consumers
        # ...

    @classmethod
    def process2_thing(cls, x):
        # methods with dependencies on other methods are classmethods so they can easily call these deps
        y = cls._common_process_thing(x)
        return .... 
```

So to follow this more obvious pattern, I recommend

1. make all of these staticmethods (I don't think there are any interdependencies?)
2. make them all `snake_case`
3. have the consumers call these static methods directly on the class, rather than inherit from a mixin and call `self.<method>()`

AlexCLeduc (approved):

PR #333, opened 2024-05-28
feat: update KCC namespace

vedantthapa:
- Move KCC resources to a separate namespace (and folder) - `configconnector-resources`
- Update instances of `cnrm-system` to `configconnector-resources`
- Remove `Namespace/cnrm-system configured` from flux alert exclusion list

I think I finally figured out the reason for #208. Both Flux and KCC CRD try to reconcile the `cnrm-system` namespace resource and end up in a race-condition, resulting in `Namespace/cnrm-system configured` spams on our slack channel.


Stephen-ONeil (approved):
:+1:

PR #338, opened 2024-06-04
fix: added data migration and field validation for trend upper and lower and other cosmetic changes

msarar:
flipped data lower and data upper for trend through data migration where lower > upper
added field validation for upper and lower for trend
fixed form field render order for trend upper and lower
added litres for trend>unit
rendered benchmarking footnotes correctly in review metadata


server/cpho/views/indicator_metadata.py
```diff
@@ -506,6 +506,26 @@ def clean_trend_segment(self):
...
+
+    def clean_data_point_upper_ci(self):
+        data_point = self.cleaned_data["data_point"]
+        data_point_upper_ci = self.cleaned_data["data_point_upper_ci"]
+        if data_point_upper_ci and data_point_upper_ci < data_point:
+            self.add_error(
+                "data_point_upper_ci",
+                tm("data_point_upper_ci_error"),
+            )
+        return data_point_upper_ci
```
AlexCLeduc, line 527: the `clean_<field>()` methods should typically only look at their own data, because other data is not guaranteed to exist. For instance, now we get 500s from line 510 is data-point is left empty

We can move these 2 checks into clean() or we can add extra checks to make sure data_point is not none

AlexCLeduc (approved):

server/cpho/migrations/0039_custom_data_migration_flip_upper_lower_trend.py
```diff
@@ -0,0 +1,64 @@
...
+        if incorrect_trend_history.count() > 0:
+            raise ValueError("Failed to flip the incorrect trend histories")
+
+    dependencies = [
+        ("cpho", "0038_remove_indicator_table_title_age_2_and_more"),
+    ]
+
+    operations = [
+        migrations.RunPython(flip_incorrect_upper_lower_trend),
+    ]
```
AlexCLeduc, line 64: Can we add a no-op so this migration is reversible, just in case?

PR #341, opened 2024-06-10
fix: revamp validation

msarar:
refactored the clean method to encompass all cleaning logic.
removed clean_<fieldname> checks as a result.

added field validation tests


AlexCLeduc (approved):
LGTM and definitely an improvement in functionality, though I feel there should be an easier way to define "required unless deleted". Maybe form logic can be generated downstream this definition? 

Maybe just a `REQUIRED_UNLESS_DELETED = [...]` list of constants on the form class? Then the fancier multi-field validation (X > Y > Z) can be done in another method that assumes they're all defined and nothing is being deleted

PR #377, opened 2024-08-21
fix: fixed export bug; French fields export

msarar:


AlexCLeduc (approved):
Woops that's a huge typo, I guess the tests were passing because of the default behaviour

PR #384, opened 2024-09-06
fix: approved data in infobase export

msarar:
This PR achieves 3 things:
1. Only Approved data included in infobase export.
2.  If indicator data/ benchmarking or trend is submitted but indicator itself is not submitted, wont export any data related to that indicator
3. if indicator, indicator data, benchmarking, trend all submitted once, but later indicator has unsubmitted changes, in this case export will show the information from the last submitted version of the indicator in all 4 worksheets


AlexCLeduc (approved):
Looks good, thanks!

PR #386, opened 2024-09-10
refactor: reduce requests

vedantthapa:
Reduce cert-manager, flux and cnpg pod's memory and cpu requests down to `112Mi` and `50m` respectively. Only exception is the `kustomize-controller`, which, seems to require more cpu than the rest so added a separate patch for it.

### DIFF
--- a/k8s/server/overlays/prod/postgres/pg-cluster.yaml
+++ b/k8s/server/overlays/prod/postgres/pg-cluster.yaml
@@ -19,7 +19,7 @@ spec:
   storage:
     storageClass: standard
     size: 2Gi
-
+  
   serviceAccountTemplate:
     metadata:
       annotations:
@@ -37,5 +37,5 @@ spec:
 
   resources:
     requests:
-      cpu: 50m
-      memory: 250Mi
+      cpu: 200m # + `isito-proxy` container = 250m
+      memory: 500Mi # + `isito-proxy` container = 612MiB
--- a/k8s/server/overlays/prod/postgres/pg-cluster.yaml
+++ b/k8s/server/overlays/prod/postgres/pg-cluster.yaml
@@ -19,7 +19,7 @@ spec:
   storage:
     storageClass: standard
     size: 2Gi
-  
+
   serviceAccountTemplate:
     metadata:
       annotations:
@@ -37,5 +37,5 @@ spec:
 
   resources:
     requests:
-      cpu: 200m # + `isito-proxy` container = 250m
-      memory: 500Mi # + `isito-proxy` container = 612MiB
+      cpu: 50m
+      memory: 250Mi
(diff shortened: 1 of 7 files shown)


k8s/server/overlays/prod/postgres/pg-cluster.yaml
```diff
@@ -37,5 +37,5 @@ spec:
 
   resources:
     requests:
-      cpu: 200m # + `isito-proxy` container = 250m
-      memory: 500Mi # + `isito-proxy` container = 612MiB
+      cpu: 50m
+      memory: 112Mi
```
Stephen-ONeil, line 41: Do the usage stats suggest reducing the postgress container's resources? I'd expect them to be occasionally spikey, and we also care much more about their availability and performance than the flux systems, etc. I'd be surprised if the app didn't suffer for reducing their resources this much, but that's just my gut!

Stephen-ONeil (changes requested):

k8s/server/overlays/prod/postgres/pg-cluster.yaml
vedantthapa: Oh, now that you mention it, there are some spikes. I was only looking at data from one of the pods from the pg-cluster for the past 2 days or so, which I've realized doesn't quite represent the reality.

For this one (or maybe for all the other ones too?), how about we keep the [resources requests as is and add limits](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#requests-and-limits) up to the old values, i.e, `250m` cpu and `512Mi`? 

Seems like this new capability on GKE doesn't require request = limits anymore.

Stephen-ONeil: Sounds good! We could also deploy to the dev cluster and try and stress test the new configuration too.

vedantthapa: Good idea! Finally a use-case for ephemeral deployment lol 😝 

I'll mark this PR as draft for now.

simardeep1792: Agreed on all points! Our main focus should be on optimizing the Flux, Kustomize, and Istio pods, which have been using more CPU and memory than necessary. A lot of resources were wasted because we couldn’t reduce the limits earlier, so refining these will help us achieve better efficiency and cost savings without compromising performance where it matters most, like with CNPG.

simardeep1792:
LGTM!

vedantthapa:
We'll tackle the stress testing issue in a separate PR. I've reverted any tuning changes related to app / db and added limits for rest of the containers, so should be good to merge this one.

Stephen-ONeil (approved):
Approved! Not sure what this failing PR linter action is, but looks like it's misconfigured so I assume someone is still mid-setup and we can ignore it. Do you know what's up there?

PR #410, opened 2025-01-16
fix: grade and hospital settings metadata

msarar:
added title and table title for grade and hospital settings with French


AlexCLeduc (approved):
Thanks!

PR #414, opened 2025-02-03
feat: remove Slack integrations

Stephen-ONeil:
Keith asked that I drop the slack integrations to simplify the architecture, to slim down anything that might get questioned if/when the project goes through the intake process to the new SSC GCP tenant. I gather the old PHAC slack is basically dead and the notification channel isn't really used any more, so I figure this is uncontentious?


msarar (approved):
LGTM. But not sure if i understand the details.
Maybe @AlexCLeduc is goi9ng to have a better idea

Stephen-ONeil:
> LGTM. But not sure if i understand the details. Maybe @AlexCLeduc is goi9ng to have a better idea

Alex said "sounds good" but also that he's on vacation and to pass the review on to you. I'll go ahead with merging and keep an eye on things, but it should be solid.

